PR #20, opened 2022-02-11
Feat: add orchestrator stack

fargito:


workspace.json
```diff
@@ -4,6 +4,8 @@
     "configuration": "packages/configuration",
     "eslint-plugin": "packages/eslint-plugin",
     "serverless-contracts-plugin": "packages/serverless-contracts-plugin",
-    "serverless-contracts": "packages/serverless-contracts"
+    "serverless-contracts": "packages/serverless-contracts",
+    "serverless-configuration": "packages/serverless-configuration",
+    "serverless-helpers": "packages/serverless-helpers"
```
guillaumeduboc, line 11: these can be ordered

packages/serverless-helpers/src/helpers/testFunctionNames.ts
```diff
@@ -0,0 +1,39 @@
+import { AWS } from '@serverless/typescript';
+
+const MAX_AWS_LAMBDA_NAME_LENGTH = 64;
+const STAGE = 'production';
```
guillaumeduboc, line 4: are we always testing with `production` as stage? 
this is considering people always use our convention of `dev`, `staging`, `production`

packages/serverless-helpers/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "module": "dist/esm/index.js",
+  "types": "dist/types/index.d.ts",
+  "scripts": {
+    "lint-fix": "yarn linter-base-config --fix",
+    "lint-fix-all": "yarn lint-fix .",
+    "linter-base-config": "eslint --ext=js,ts .",
+    "package": "rm -rf dist && yarn package:cjs && yarn package:esm && yarn package:types",
+    "package:cjs": "NODE_ENV=cjs yarn transpile --out-dir dist/cjs --source-maps",
+    "package:esm": "NODE_ENV=esm yarn transpile --out-dir dist/esm --source-maps",
+    "package:types": "ttsc -p tsconfig.build.json",
```
guillaumeduboc: don't forget to change `package-...`

packages/serverless-helpers/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "test": "yarn test-linter && yarn test-type && yarn test-unit",
+    "test-linter": "yarn linter-base-config .",
+    "test-type": "tsc --noEmit --emitDeclarationOnly false",
+    "test-unit": "jest --group=unit --runInBand --collectCoverage --logHeapUsage --passWithNoTests",
+    "transpile": "babel src --extensions .ts --quiet",
+    "watch": "rm -rf dist && concurrently 'yarn:package:* --watch'"
```
guillaumeduboc: and here `package-*`

backend/orchestrator/project.json
```diff
@@ -0,0 +1,102 @@
+{
```
guillaumeduboc, line 1: in `swarmion/template` we decided to only define targets in `nx.json` and not in `project.json` anymore
if we want our backend to be independent from the rest it makes sense, but I find it weird having both conventions

fargito: This is fixed in a separate commit

packages/serverless-helpers/package.json
fargito: This is fixed in a separate commit

backend/orchestrator/resources/dynamodb.ts
```diff
@@ -0,0 +1,23 @@
+export const PARTITION_KEY = 'pk';
+export const SORT_KEY = 'sk';
+
+export const OrchestratorTable = {
```
adriencaccia, line 4: We can leverage lift's [Database - DynamoDB Single Table construct](https://github.com/getlift/lift/blob/master/docs/database-dynamodb-single-table.md) to have less boilerplate and better DDB configuration out of the box

workspace.json
fargito: I also reordered the vscode workspaces

backend/orchestrator/libs/models/serviceEvent.ts
```diff
@@ -0,0 +1,32 @@
...
+      }) => `${applicationId}#SERVICE#${serviceId}`,
+    },
+    [SORT_KEY]: {
+      sortKey: true,
+      hidden: true,
+      default: ({ timestamp }: { timestamp: string }) => `EVENT#${timestamp}`,
+    },
+    serviceId: { type: 'string', required: true },
+    applicationId: { type: 'string', required: true },
+    timestamp: { type: 'string', required: true },
```
adriencaccia, line 27: Instead of maintaining the timestamp ourselves, we can leverage [dynamodb-toolbox `created` attribute](https://github.com/jeremydaly/dynamodb-toolbox#specifying-entity-definitions)

packages/serverless-helpers/src/helpers/testFunctionNames.ts
fargito: Hmm that is a good point. However maybe we should open a separate issue on this, what do you think?

backend/orchestrator/functions/requestSyncDeployment/handler.ts
```diff
@@ -0,0 +1,19 @@
+import { requestSyncDeployment } from '@swarmion/orchestrator-contracts';
+import { applyHttpMiddlewares } from '@swarmion/serverless-helpers';
+
+import ServiceEventEntity from 'libs/models/serviceEvent';
+
+const handler = requestSyncDeployment.handler(async event => {
+  const { serviceId, applicationId } = event.body;
+
+  const timestamp = new Date().toISOString();
```
adriencaccia, line 9: Per my previous comment, we would not need to manually create this

packages/serverless-helpers/src/helpers/testFunctionNames.ts
adriencaccia: This test is here to cover the worst case scenario (with the longest name), so we use `production` as it is the one with the most letters

adriencaccia (changes requested):

backend/orchestrator/functions/requestSyncDeployment/handler.ts
fargito: I replaced it by an ulid which is stronger than a timestamp

backend/orchestrator/libs/models/serviceEvent.ts
fargito: That is right, I replaced it with an `eventId` that is an ulid

packages/serverless-helpers/src/helpers/testFunctionNames.ts
adriencaccia: What we can do is add a comment to explain that in the file, WDYT ?

fargito: I added it as a jsdoc so it also available in service tests

backend/orchestrator/resources/dynamodb.ts
fargito: I think we can use a more fine-grained permissions system for this table, by default Lift grants all permissions to all lambdas https://github.com/getlift/lift/blob/master/docs/database-dynamodb-single-table.md#permissions

adriencaccia (approved):

PR #27, opened 2022-03-10
chore(contracts): enable tree-shaking in contracts

fargito:


adriencaccia (approved):

adriencaccia (changes requested):
We can add the property to all packages

adriencaccia (approved):

PR #33, opened 2022-03-16
chore: setup vscode workspace

fargito:


adriencaccia (changes requested):
Nice !
There is one last thing to do: add the common `.vscode` folder to `commonConfiguration` and symlink it to each package

commonConfiguration/.vscode/launch.json
```diff
@@ -0,0 +1,47 @@
...
+        "functions/${input:functionName}/handler.mock.json"
+      ],
+      "sourceMaps": true,
+      "smartStep": true,
+      "outFiles": ["**/.esbuild/**/*.js"],
+      "protocol": "inspector",
+      "autoAttachChildProcesses": true,
+      "console": "integratedTerminal",
+      "outputCapture": "console"
+    },
```
adriencaccia, line 34: We can remove the inputs and the first configuration, since there are no lambdas in the repo

adriencaccia (changes requested):

commonConfiguration/.vscode/launch.json
fargito: This will be useful in the future, when we add the orchestrator backend

adriencaccia: fair enough

PR #53, opened 2022-04-01
fix(nx): make nx command work again

adriencaccia:
Release https://github.com/nrwl/nx/releases/tag/13.9.3 remove the nx cli
from @nrwl/cli.
We need to explicitly add the nx dependency now.


fargito (approved):
:ok_hand:

PR #84, opened 2022-04-08
chore: update tsconfig

fargito:


tsconfig.json
```diff
@@ -10,11 +10,8 @@
     "noUnusedParameters": true,
     "sourceMap": true,
     "strict": true,
-    "noImplicitReturns": true,
```
adriencaccia, line 13: We better leave this enabled, as it is not enabled by default: https://www.typescriptlang.org/tsconfig#noImplicitReturns

fargito: Ok!

PR #88, opened 2022-04-11
Feat(nx-plugin) : resolve dependencies after generation

guillaumeduboc:


guillaumeduboc:
Do you think using the `--cached` option with `yarn add` would be relevant here?

packages/nx-plugin/src/generators/helpers/index.ts
```diff
@@ -1,4 +1,6 @@
 export * from './createFiles';
 export * from './normalizeOptions';
 export * from './packageGenerator';
+export * from './symlink';
+export * from './udatePackages';
```
fargito, line 5:
```suggestion
export * from './updatePackages';
```

fargito:
This looks promising!

guillaumeduboc:
With `--cached` option, this is the resolved `package.json`
Capture d’écran 2022-04-15 à 18 26 28

MaximeVivier:
For the generation of the package.json, we decided not to go with the `latest` version for each package. It causes the installation of new versions which implies that the developer needs to test new version.
We don’t want that, we prefer using the latest version used in the whole project for each package.
The `--cached` tag on the `yarn add` command allow us to do that. it works but the version defined is not the one expected. Indeed, in the `yarn.lock` file the version from the package.json is listed as being resolved by the highest version used in the project as expected because it has a carret, that’s why it works and that’s what we want. But the problem is that the lower version with the carret in the package.json may mislead the developer.
In the end, it works but it’s not optimal for the DevX.

fargito (approved):
This is awesome! Thanks a lot @guillaumeduboc and @MaximeVivier for this :heart_eyes:

PR #89, opened 2022-04-11
Refactor: replace cloudformation by cdk

Sc0ra:


backend/orchestrator/tsconfig.json
```diff
@@ -3,7 +3,12 @@
   "compilerOptions": {
     "preserveSymlinks": true,
     "baseUrl": ".",
-    "esModuleInterop": true
+    "esModuleInterop": true,
+    "paths": {
+      "functions/*": ["functions/*"],
+      "libs/*": ["libs/*"],
+      "resources/*": ["resources/*"]
+    }
```
adriencaccia, line 11: It is already infered from `baseUrl`, we do not neet to explicitely add the `paths` here

backend/orchestrator/libs/orchestratorTable.ts
```diff
@@ -1,19 +1,17 @@
...
-const OrchestratorTable = new Table({
-  name: getEnvVariable('ORCHESTRATOR_TABLE_NAME'),
+export const OrchestratorTable = new Table({
+  name: ORCHESTRATOR_TABLE_NAME,
   partitionKey: PARTITION_KEY,
   sortKey: SORT_KEY,
   autoExecute: true,
   autoParse: true,
   DocumentClient,
 });
```
adriencaccia, line 17: We can leave the `default` export and rename the file into `OrchestratorTable.ts` (with a capital O).
The convention is when a file is exporting is single value, we use a default export with the same name as the file.

adriencaccia (changes requested):

fargito (approved):

PR #107, opened 2022-04-29
feat: adding eslint rule for aws-sdk v2 imports

### LINKED ISSUE #106: Add eslint rule to warn on import of aws-sdk root
Just like `lodash`, exports from `aws-sdk` must not be imported from its root, but from a specific path, such as`aws-sdk/clients/.....`

This is because `aws-sdk` is included as cjs in the lambda runtime but cannot be tree-shaked. This is a big performance issue.

Eikix:
Resolves #106 

Hi @fargito ! Here's a small PR to solve aws-sdk v2 imports issue ✨ 🚀 .

have a good one!

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -77,6 +77,10 @@ module.exports = {
             message: 'Please use lodash/{module} import instead',
           },
           {
+            name: 'aws-sdk',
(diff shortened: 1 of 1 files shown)


fargito:
@allcontributors please add @Eikix for code

PR #111, opened 2022-04-29
Functionnal contracts

Sc0ra:


backend/orchestrator/functions/requestSyncDeployment/handler.ts
```diff
@@ -1,21 +1,22 @@
...
   await ServiceEventEntity.put({ serviceId, applicationId, eventId });
 
   return { status: 'ACCEPTED', message: 'processing' };
-});
+};
 
 export const main = applyHttpMiddlewares(handler, {
-  inputSchema: requestSyncDeployment.inputSchema,
+  inputSchema: getInputSchema(requestSyncDeployment),
   outputSchema: requestSyncDeployment.outputSchema,
```
fargito, line 21: This introduces a disymmetry between the way `inputSchema` and `outputSchema` are handled, is there a way to unify this?

packages/serverless-contracts-plugin/src/plugin/utils/generateOpenApiDocumentation.ts
```diff
@@ -12,8 +15,10 @@ export const generateOpenApiDocumentation = (serverless: Serverless): void => {
     .contracts as ServerlessContracts;
 
   const contractDocumentations = Object.values(provides)
-    .filter(isInstanceOfDocumentedContract)
-    .map(contract => contract.openApiDocumentation);
+    .map(contract => getOpenApiDocumentation(contract))
+    .filter(
+      documentation => documentation !== undefined,
+    ) as ContractOpenApiDocumentation[];
```
fargito, line 21: Is this cast necessary?

fargito: Maybe we could use a type guard instead

packages/serverless-contracts/src/contracts/apiGateway/__tests__/httpApiContract.test.ts
```diff
@@ -1,5 +1,12 @@
 /* eslint-disable max-lines */
 import { ApiGatewayContract } from '../apiGatewayContract';
+import {
+  getCompleteTrigger,
+  getFullContractSchema,
+  getInputSchema,
+  getOpenApiDocumentation,
+  getTrigger,
+} from '../features';
```
fargito, line 9: Maybe we could benefit from the `features` split to also test each feature independently in separate test files

packages/serverless-contracts/src/contracts/apiGateway/features/lambdaTrigger.ts
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Returns an enhanced serverless function trigger associated to an ApiGatewayContract
+ *
+ * @param contract your ApiGatewayContract
+ * @param additionalConfig for example an authorizer reference, ...
+ */
+export const getCompleteTrigger = <Contract extends ApiGatewayContract>(
+  contract: Contract,
+  additionalConfig: ApiGatewayLambdaConfigType<
```
fargito, line 33: These two functions now seem very similar and could maybe be merged in a single `getTrigger` function with an optional `additionalConfig` argument

packages/serverless-contracts/src/contracts/apiGateway/types/index.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Computed request parameters. This enables the call to the contract to be type-safe
+ */
+export interface RequestParameters<BodyType> {
+  method: HttpMethod;
+  path: string;
+  body?: BodyType;
+  headers?: Record<string, string>;
+  queryStringParameters?: Record<string, string>;
+}
+
```
fargito, line 20: Index files should only contain exports, can you move these types to a separate file?

packages/serverless-contracts/src/features/fullSchema.ts
```diff
@@ -0,0 +1,19 @@
...
+  contract: ServerlessContract,
+): JSONSchema => {
+  switch (contract.contractType) {
+    case 'apiGateway': {
+      // @ts-expect-error FIXME: problem with types here
+      return getApiGatewayFullContractSchema(contract);
+    }
+    case 'cloudFormation': {
+      return getCloudFormationFullContractSchema(contract);
+    }
```
fargito, line 17: Can you add an exhaustiveness check here? https://kimsereylam.com/typescript/2022/02/11/exhaustiveness-checking-in-typescript.html

user-docs/documentation/docs/contracts/api-gateway.md
```diff
@@ -148,32 +148,36 @@ in order to validate the input and/or the output of your lambda.
 
 ### Type the lambda input and output
 
-On the handler side, you can use the `handler` method on the contract to correctly infer the input and output types from the schema.
+On the handler side, you can use the `HandlerType` generic type on the contract to correctly infer the input and output types from the schema.
 
 ```ts
-const handler = myContract.handler(async event => {
+const handler: HandlerType<typeof myContract> = async event => {
   event.pathParameters.userId; // will have type 'string'
```
fargito, line 155: Since we are most likely going to implement parsing and validation in Swarmion, it may be a good idea to keep a wrapper function. It could look something like 
```
const handler = contractHandler(myContract)(async event => {
  event.pathParameters.userId; // will have type 'string'
  ...
});
 ```

File: packages/serverless-contracts/src/features/fullSchema.ts (hunk 6 above)
Comment (Sc0ra) [reply packages/serverless-contracts/src/features/fullSchema.ts:17]: Typescript already throw an error if i remove one of the case
If i had an exhaustive check, the typing of the function will be modified because the default case will not be able to return a fullContractSchema

File: packages/serverless-contracts/src/contracts/apiGateway/features/axiosRequest.ts [hunk 8]
```diff
@@ -0,0 +1,23 @@
...
+import { OutputType, RequestArguments } from '../types';
+import { getRequestParameters } from './requestParameters';
+
+export const getAxiosRequest = async <Contract extends ApiGatewayContract>(
+  contract: Contract,
+  axiosClient: AxiosInstance,
+  requestArguments: RequestArguments<Contract>,
+): Promise<AxiosResponse<OutputType<Contract>>> => {
+  const { path, method, queryStringParameters, body, headers } =
+    // @ts-ignore: Type instantiation is excessively deep and possibly infinite.
```
fargito, line 13: Hmm this probably means that our types are not efficient. We can look at this together

packages/serverless-contracts-plugin/src/plugin/utils/generateOpenApiDocumentation.ts
```diff
@@ -1,19 +1,21 @@
 import { OpenAPIV3 } from 'openapi-types';
 import Serverless from 'serverless';
 
-import { isInstanceOfDocumentedContract } from '@swarmion/serverless-contracts';
+import { getOpenApiDocumentation } from '@swarmion/serverless-contracts';
 
 import { ServerlessContracts } from 'types/serviceOptions';
 
+import { isDefined } from './isDefined';
```
fargito, line 8: We can use lodash for this

packages/serverless-contracts/src/features/fullSchema.ts
fargito: Great then!

user-docs/documentation/docs/contracts/api-gateway.md
```diff
@@ -148,10 +148,10 @@ in order to validate the input and/or the output of your lambda.
 
 ### Type the lambda input and output
 
-On the handler side, you can use the `handler` method on the contract to correctly infer the input and output types from the schema.
+On the handler side, you can use the `getLambdaHandler` function on the contract to correctly infer the input and output types from the schema.
 
 ```ts
-const handler = myContract.handler(async event => {
+const handler: getLambdaHandler(myContract)(async event => {
```
fargito, line 154:
```suggestion
const handler = getLambdaHandler(myContract)(async event => {
```

fargito:
A most welcome refactor ! A few minor changes and it's all good 👍🏻

packages/serverless-contracts/src/contracts/apiGateway/features/axiosRequest.ts
```diff
@@ -0,0 +1,23 @@
...
+    getRequestParameters(contract, requestArguments);
```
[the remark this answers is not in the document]
fargito: This way we don't have a double type inference

packages/serverless-contracts-plugin/src/plugin/utils/generateOpenApiDocumentation.ts
Sc0ra: It doesn't seem to work for now, so I think we can keep it that way

fargito (approved):
Thanks a lot!

PR #117, opened 2022-05-13
Docs/contracts reference

Sc0ra:


user-docs/documentation/docusaurus.config.js
```diff
@@ -74,6 +80,26 @@ const config = {
...
+            path: 'packages/serverless-contracts',
+            entry: {
+              index: 'src/index.ts',
+            },
+          },
+        ],
+        minimal: true,
+        readmes: true,
+        debug: true,
+        tsconfigName: 'tsconfig.docs.json',
```
fargito, line 99: I think it would be better to place this config in the `serverless-contracts` package, what do you think?

PR #134, opened 2022-05-27
feat(user-docs): add stage params for static website construct

fargito:
This PR add per-stage support to deploy documentation to cloudfront. The issue here is that for some environments we need empty variables and for the moment it is not possible with Serverless.

I'll check what we can do


fargito:
Waiting for https://github.com/getlift/lift/pull/217

PR #157, opened 2022-06-16
fix: fix linter-base-config script

Thomasprelot:
The command `"linter-base-config" : "eslint --ext=js,ts ."` is linting every file with the appropriate extension in the current directory.

This is not the expected behaviour since the command `yarn lint-fix-all` serves this purpose.
Especially, when this command is triggered by `lint-staged` (eg in pre-commit), linter is applied to all files, not only the staged files, which can cause unwanted warnings/errors (especially in pre-commit).

The fix is the following update in the command : `"linter-base-config" : "eslint --ext=js,ts"`


adriencaccia (changes requested):
Thanks for the PR @Thomasprelot!

Ok for the changes :ok_hand: 

Just one thing, can you change the following line to run the script `lint-fix-all` instead of `lint-fix` at the root package ?
 https://github.com/swarmion/swarmion/blob/07d0c848a0008e57899ed06d2d41d6e532280fb9/package.json#L29

Indeed, it is `lint-fix-all` that should be run

adriencaccia (approved):

adriencaccia:
@allcontributors please add @Thomasprelot  for code

Issue #37, opened 2022-03-18
Allow using node-fetch on ApiGateway contracts


ThomasAribart:
I think `node-fetch` is more standard as the `fetch` already available in browsers.

```typescript
import fetch from "node-fetch" // <= Or nothing in browser

const response = await fetch(
  baseUrl,
  myApiGatewayContract.getFetchArgs(event)
)

// alternatively
const response = await myApiGatewayContract.fetch(
  baseUrl,
  event
)
```

PR #159, opened 2022-06-17
feat(contracts): add getFetchRequest feature for api gateway contract

Sc0ra:
Should close https://github.com/swarmion/swarmion/issues/37


packages/serverless-contracts/package.json
```diff
@@ -40,6 +40,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.18.3",
+    "cross-fetch": "^3.1.5",
```
fargito, line 43: We can move this to devDependencies I think since it only used for typing

Sc0ra: By referencing browser fetch module, i was able to remove cross-fetch deps entirely

fargito (approved):
Great! Thanks a lot

Issue #131, opened 2022-05-24
[RFC] Improve Swarmion documentation


fargito:
In order to provide a useful tool to developers, we need to clearly explain our architectural choices and guide users.

I have drafted a first structure, this is open for suggestions!

## Tutorials

- Generate your first Swarmion app
- Get started on AWS
- Add your first Serverless service
- Add your first shared Typescript library
- Add your first frontend
- Create your first contract between the frontend and the backend

## How-to guides

- Add authentication to your Serverless contracts
- Managing package upgrades with Swarmion
- Generating libraries with Swarmion
- Generating Serverless services with Swarmion
- Migrating from a monolith to Swarmion
- Share variables between your services
- Configuring your CI/CD
    - Github actions
    - CircleCI
    - GitlabCI

## Explanation

- Why use a microservices approach?
- Why use a monorepo?
- Why use yarn3 and workspaces?
- Why share code between services?
- Why build Typescript packages?
- Serverless bundling → from TS to JS (why esbuild)
- Why use Nx as a monorepo management tool?
- Understanding configuration composition
    - Typescript
    - Eslint
    - Jest
    - linstaged
- Understanding the Swarmion VSCode configuration
- Why check for circular references?
- Why use a precommit hook?
- Understanding Serverless contracts
    - Types of contracts
    - Concepts
    - Why are they using functionnal programming?
- Miscellaneous tools
    - Why Fresko?
    - Why synckpack?
    - Why Commitlint?

## References

- `@swarmion/serverless-contracts`
- `@swarmion/nx-plugin`
- `@swarmion/serverless-plugin`
- `@swarmion/nx-plugin`
- Code structure
    - Monorepo structure
    - Root config files
    - Serverless service structure
    - Package structure
    - Frontend shared package structure
    - Frontend structure

LouisPinsard:
Hello @adriencaccia @fargito,

Great job with Swarmion so far I am a huge (a bit biased) fan !

I think it is important to add some context on the core service that holds the httpApi.
Indeed, almost all services will implicitly depend on this services output. I think it is quite important to provide a bit of guidelines on what you should put (or not) in this service as it breaks the independence of the deployments.

In my opinion in a classical AWS serverless stack it should only contain:
- shared DynamoDB database
- shared eventbridge
- shared api gateway
- if you use Cognito JWT authorizer, it's easier to put it in core also

And that's it (no lambda). It should only contain resources that won't change much while the project lives.

[...]

PR #161, opened 2022-06-17
docs(user-docs): new structure

fargito:
See https://github.com/swarmion/swarmion/issues/131


user-docs/documentation/how-to/6-configure-ci-cd.md
```diff
@@ -3,3 +3,198 @@ sidebar_position: 6
 ---
 
 # Configure your CI/CD
+
+## Choose the correct IAM policy to give your CI user
```
adriencaccia, line 7: If using github actions, we can recommend to follow these [guidelines to setup OIDC](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services), removing the need to store long lived credentials, and this removing the need to specify a policy (we can give admin access)

adriencaccia (approved):

PR #182, opened 2022-07-01
Feat(create-swarmion-app): download version of the example corresponding to the package

### LINKED ISSUE #175: Use versionned templates in `create-swarmion-app`
Currently, `create-swarmion-app` fetches the `examples` folder on its latest version (`main`). However, this is risky since changes can be made on main that have not yet been validated.

In order to fix this issue, we should get the current version in `create-swarmion-app` and use it to build the remote url in create-`packages/swarmion-app/src/create-app.ts`

guillaumeduboc:
resolves #175


packages/create-swarmion-app/src/helpers/examples.ts
```diff
@@ -28,6 +28,14 @@ const isUrlOk = async (url: string): Promise<boolean> => {
   }
 };
 
+const getFileNameFromBranch = (branch: string): string => {
+  if (branch.match(/v(\d).(\d).(\d)/)) {
```
fargito, line 32: You may encounted issues here with versions that have more than two digits. For example v1.12.488 is a valid version number but will not be matched

fargito: Also we may want to have alpha or beta releases, how will this be handled ?

guillaumeduboc: Very good point

guillaumeduboc: I added tests to cover these use-cases 😄

fargito (approved):
Thanks @guillaumeduboc !

PR #186, opened 2022-07-01
feat: add a cdk-bridge autocompletion fn for class attributes

Eikix:
chore: move helper function out of plugin code

chore: fix preemptively esm / mjs module import bugs

feat: add variable name helper tests


Eikix:
wrong usage of fork main branch to submit PR

PR #209, opened 2022-07-20
Fix(nx-plugin) : add `serverless-iam-roles-per-function` dependency to services generated

### LINKED ISSUE #206: [Generators] `serverless-iam-roles-per-function` is referenced in `serverless.ts` but not installed
Steps to reproduce:
- Start from a fresh install of Swarmion
- Run  `yarn generate-service my-service`
- Then `yarn deploy`

The deployment of the generated package fails with error:
```
       Error:
       Serverless plugin "serverless-iam-roles-per-function" not found. Make sure it's installed and listed in the "plugins" section of your serverless config file. Run "serverless plugin install -n serverless-iam-roles-per-function" to install it.
```

We need to add it in the service package.json during the generation

guillaumeduboc:
fixes #206


fargito (approved):
Thanks for this!

PR #222, opened 2022-07-25
chore: add netlify docs preview

adriencaccia:


fargito (approved):
Nice! This is going to be so much easier

PR #239, opened 2022-08-26
fix(contracts): prevent using URL class to build fetch path with no base url

### LINKED ISSUE #208: [FetchRequest] `getFetchRequest` doesn't allow undefined `baseUrl`
At the moment, although it is not typed as required, `getFetchRequest` requires a `baseUrl` in its `options` argument : providing an empty object for `options` results in the following error : `Failed to construct 'URL': Invalid URL`.

This is because `getFetchRequest` creates a new URL and provide it to the `fetchFunction`. In the absence of the `baseUrl` option, it should only call the `fetchFunction` with the contract `path` as first parameter (the `fetchFunction` will then be responsible of adding the `baseUrl`).

Also, the `options` argument is always required in `getFetchRequest`. It could be optional and set to an empty object by default if no option is required.

Sc0ra:
Fixes #208 
The problem came from the fact that new URL validate the url format and require it to start with http:// etc
So now, if baseUrl is not defined, we don't use the URL class and we concatenate the path and the query params manually

### DIFF
--- a/packages/serverless-contracts/src/contracts/apiGateway/__tests__/fetchRequest.test.ts
+++ b/packages/serverless-contracts/src/contracts/apiGateway/__tests__/fetchRequest.test.ts
@@ -127,4 +127,36 @@ describe('apiGateway fetch request', () => {
       );
     });
   });
+
+  describe('httpApi without base url', () => {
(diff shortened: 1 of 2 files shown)


fargito (approved):
Great!

PR #243, opened 2022-08-31
fix(serverless-helpers): use region to validate function names

fargito:
Currently the test is invalid since the blocking item is never the function name but rather the role name


adriencaccia:
I created the following plugin https://github.com/adriencaccia/serverless-custom-iam-roles-per-function to solve the problem of lambda role length.

I think we should keep the test on functions names length, and add the plugin to swarmion 😉

fargito:
Ok sounds fair enough

fargito:
See https://github.com/swarmion/swarmion/pull/244

PR #250, opened 2022-09-05
Doc: add info on header case in http/2

Yoctoboy:
Hey guys,

As per HTTP/2 spec, requests' headers must be lowercase. However, since type-inference is case-sensitive in typescript, we may encouter errors when having a header defined like this (with uppercase letter(s)):
Screenshot from 2022-09-05 14-53-43

Here, when using `getHttpLambdaHandler` from a contract with such headers, ajv looks for a `myHeader` key but in HTTP/2, curl defaults to lowercasing all headers which leads to a validation error.

This is just a proposition as swarmion's doc is not supposed to be about http standards, but I thought this example was a bit misleading.


adriencaccia:
LGTM, thanks for the contribution 😉 
I will let @fargito approve

PR #265, opened 2022-09-11
Switch to vitest ⚡

adriencaccia:
Everything is working great 🚀 

One thing that is still missing though, is the ability to debug test with the https://github.com/vitest-dev/vscode extension.

I opened an issue https://github.com/vitest-dev/vscode/issues/84 and a pull request https://github.com/vitest-dev/vscode/issues/83 to solve it.

I think we can still merge and release this without waiting for my fix to be released.


backend/orchestrator/package.json
```diff
@@ -18,8 +18,8 @@
     "test": "yarn test-linter && yarn test-type && yarn test-unit && yarn test-circular",
     "test-circular": "yarn depcruise --validate .dependency-cruiser.js .",
     "test-linter": "yarn linter-base-config .",
-    "test-type": "tsc --noEmit",
-    "test-unit": "jest --group=unit --runInBand --collectCoverage --logHeapUsage --passWithNoTests"
+    "test-type": "tsc --noEmit --emitDeclarationOnly false",
```
fargito, line 21: Shouldn't this be in the `tsconfig.json`?

fargito (approved):
Massive job!

PR #283, opened 2022-10-07
fix: add undefined queryStringParameters values management in getFetchRequest

### LINKED ISSUE #271: Undefined query params management in getFetchRequest
Hello swarmion team,

When undefined queryStringParameters are given to getFetchRequest, they are not ignored.

For example:

``` typescript
getFetchRequest(getScenariosForListContract, authFetch, {
          queryStringParameters:  { pageToken: undefined },
        })
```

results in the following call: 
**/scenarios?pageToken=undefined**

Wouldn't it be better to ignore undefined values and call the following route:
**/scenarios** 

Thanks for the great work!

StanHannebelle:
This adds management of undefined queryStringParameters values in getFetchRequest and resolves #271

### DIFF
--- a/packages/serverless-contracts/src/contracts/apiGateway/__tests__/fetchRequest.test.ts
+++ b/packages/serverless-contracts/src/contracts/apiGateway/__tests__/fetchRequest.test.ts
@@ -21,7 +21,10 @@ describe('apiGateway fetch request', () => {
 
   const queryStringParametersSchema = {
     type: 'object',
-    properties: { testId: { type: 'string' } },
+    properties: {
+      testId: { type: 'string' },
+      optionalParam: { type: 'string' },
+    },
     required: ['testId'],
     additionalProperties: false,
   } as const;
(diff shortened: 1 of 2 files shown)


packages/serverless-contracts/src/contracts/apiGateway/__tests__/fetchRequest.test.ts
```diff
@@ -95,6 +94,38 @@ describe('apiGateway fetch request', () => {
...
+        httpApiContract,
+        mockedFetch as unknown as typeof fetch,
+        {
+          pathParameters: {
+            userId: 'azer',
+            pageNumber: 'zert',
+          },
+          queryStringParameters: {
+            testId: undefined,
+          },
```
adriencaccia, line 109: To better test the functionality, let's add another property with a defined value

adriencaccia (changes requested):

packages/serverless-contracts/src/contracts/apiGateway/__tests__/fetchRequest.test.ts
StanHannebelle: Just updated the test :)

adriencaccia (approved):

PR #287, opened 2022-10-11
chore(ci): update actions/cache and actions/checkout to v3

fargito:


adriencaccia:
Are there any breaking changes?

fargito:
No, I don't think so

fargito:
It simply uses Node16 instead of Node12

PR #295, opened 2022-10-14
Expose api gateway event mocks

adriencaccia:
- feat(serverless-helpers): add api gateway mocks
- refactor(serverless-contracts): use mock helpers from serverless-helpers
- test(serverless-helpers): add handler unit test


packages/serverless-helpers/src/mocks/apiGateway.ts
```diff
@@ -38,7 +38,7 @@ export const getRequestContextMock = (
   ...args,
 });
 
-export const getRequestContextMockV2 = (
+export const getAPIGatewayEventRequestContextMockV2 = (
```
fargito, line 41:
```suggestion
export const getAPIGatewayV2EventRequestContextMock = (
```
Wouldn't it be a better naming?

adriencaccia: Yep, changed it

fargito (approved):

fargito:
@adriencaccia do you think this PR could be merged? It would nicely integrate with #358

fargito:
@adriencaccia I rebased it on main but it seems to have a problem with tests.

I was event thinking, wouldn't it be great to expose a serverless-contracts feature called `buildTestEvent` to bring even more type-safety to tests?

It could be used like:
```
// handler.test.ts

const { event, context, callback } = buildHandlerTestEvent(myContract)({
    // type-safe values from the contract...
});

const result = await main(event, context, callback);

expect(result.xxxx).toBe('yyyy') // whatever assertion on the result

```

Therefore we would remove the need to directly instantiate `getAPIGatewayV2EventRequestContextMock` or other mock builders. This function would directly populate the event context in the test event.

WDYT?

adriencaccia:
> @adriencaccia do you think this PR could be merged? It would nicely integrate with #358

Yeah, I'll look into it this afternoon

adriencaccia (commented):
LGTM 🎉 
I cannot approve the PR since I am the one that created it 😅
You can merge it

fargito:
Great! Thanks

PR #312, opened 2022-10-21
fix: replace queryStringParameters undefined by empty object

bchangal:
- @LouisPinsard: Discussed with @Sc0ra about it pull request this implementation seems easier.

 What do you think @fargito  ?

We didn't yet tested deeply the implementation proposal.
 
Would be happy to implement if needed !


fargito (approved):
Thanks a lot @bchangal.

It is indeed a lot simpler this way and overrides the strange way with which ApiGateway handles its query parameters.

I'll take the time to properly test it before merging though.

fargito:
@allcontributors please add @bchangal for code

PR #317, opened 2022-10-25
fix mui import in muiTheme.ts

Yoctoboy:
## Problem

A weird `createTheme_default is not a function` appears when importing `createTheme` from `@mui/material/styles`.

The problem is known by the community and has a simple fix, which is in this PR

## Links
- Issue on material-ui repo : https://github.com/mui/material-ui/issues/31835
  - (N.B.: OP thinks this is an "issue with esbuild" but it is also reproducible with vite)
- Comment suggesting fix : https://github.com/mui/material-ui/issues/31835#issuecomment-1147055123

I checked that there is no other such import in the swarmion repo.


fargito:
@allcontributors please add @Yoctoboy for code

PR #335, opened 2022-11-03
chore: keep comments in .d.ts files

fargito:
We currently remove the comments from our .d.ts files.

This is a problem for our published libraries as they do not publish their nice JSDoc.

Also for the two starters, it seems to me that the benefits of removing comments is non-existent since we do not use `tsc` to transpile from TS to JS. On the contrary, keeping them would improve internal documentation.

Check out https://github.com/microsoft/TypeScript/issues/14619 for more context.

WDYT?


adriencaccia (approved):
LGTM, it is a nice addition!
Thanks @fargito

PR #347, opened 2022-11-18
Use tsup to transpile

GuillaumeLagrange:
- feat(full-stack-example): use tsup to bundle frontend/shared
- feat(orchestrator-contracts): use tsup to transpile
- feat(cdk-plugin): use tsup to transpile
- chore(nx-plugin): set @nrwl/devkit as dependency
- feat(nx-plugin): use tsup to transpile
- feat(eslint-plugin): use tsup to transpile
- feat(serverless-configuration): use tsup to transpile
- feat(serverless-contracts): use tsup to transpile
- feat(serverless-contracts-plugin): use tsup to transpile
- feat(serverless-helpers): use tsup to transpile
- chore: remove babel config files


fargito (approved):
LGTM

adriencaccia (changes requested):
Great work @GuillaumeLagrange 💪 

However, the examples are not fully migrated: 
image
And the generators have not been updated to use `tsup` instead of `babel`.

I think it should be done before merging this.

WDYT?

packages/nx-plugin/package.json
```diff
@@ -49,7 +49,8 @@
   },
   "dependencies": {
     "@nrwl/devkit": "^15.3.0",
-    "@nrwl/linter": "^15.3.0"
+    "@nrwl/linter": "^15.3.0",
+    "typescript": "^4.9.4"
```
adriencaccia, line 53: This might be a bit too greedy to make the users download this specific version of `typescript`.
Even more that most likely no one will ever the types of the package since it will be used by `nx`.

Isn't there a way to get around with `devDevpendencies` and/or `peerDependencies`?

adriencaccia (changes requested):

packages/eslint-plugin/package.json
```diff
@@ -40,6 +40,7 @@
     "watch": "pnpm clean && concurrently 'pnpm:package-* --watch'"
   },
   "dependencies": {
+    "eslint": "^8.29.0",
```
guillaumeduboc, line 43: I don't think we should have eslint as dependency

packages/serverless-cdk-plugin/package.json
```diff
@@ -40,7 +40,11 @@
   "dependencies": {
     "@serverless/typescript": "^3.25.0",
     "@types/serverless": "^3.12.8",
+    "aws-cdk-lib": "^2.54.0",
+    "constructs": "^10.1.187",
+    "json-schema-to-ts": "^2.6.2",
     "lodash": "^4.17.21",
+    "serverless": "^3.25.1",
```
guillaumeduboc, line 47: they are already in peerDependencies
eslint should be configured to accept that

guillaumeduboc:
I looked at the last commit and I suggest https://github.com/swarmion/swarmion/commit/5b2d0dc4bf90a41f53e27161e0b335b70a12a73c instead

- for devDependencies only used for types, use `import type ...` 
- for libraries that have peerDependencies, override the eslint rule
- clean up cdk plugin dependencies that should be peerDeps
- add eslint rule to examples

packages/nx-plugin/package.json
GuillaumeLagrange: done

adriencaccia (approved):

Issue #345, opened 2022-11-18
Replace Serverless Framework with CDK


fargito:
## Migration plan

- We plan to make CDK the default IaC tool for deploying Serverless micro-services in Swarmion
- The goal is to have a starter usable in the next few weeks, the work is tracked in #351 
- Then during the first few months of January, we will go and implement the existing serverless contracts triggers on CDK, to have the same support on both frameworks

But we do not plan to deprecate the Serverless framework anytime soon 😃
Indeed, a lof (if not all) of our users are using SF, so we will keep supporting it!

adriencaccia:
Hey @tainmar, we need to add more information on this issue 😅 

Until that, this is where things stand:
- We plan to make `CDK` the default IaC tool for deploying Serverless micro-services in Swarmion
- The goal is to have a starter usable in the next few weeks
- Then during the first few months of January, we will go and implement the existing serverless contracts triggers on CDK, to have the same support on both frameworks

But **we do not plan to deprecate the Serverless framework anytime soon** 😃 
Indeed, a lof (if not all) of our users are using SF, so we will keep supporting it!

PR #351, opened 2022-11-21
Add an example with CDK

adriencaccia:
This pull request if the first part of #345.


examples/swarmion-with-cdk/backend/core/resources/stack.ts
```diff
@@ -13,6 +14,18 @@ export class CoreStack extends Stack {
...
+
+    const coreApi = new RestApi(this, 'CoreApi', {
+      // the stage of the API is the same as the stage of the stack
+      description: `Core API - ${stage}`,
+      deployOptions: {
+        stageName: stage,
+      },
+    });
+
+    const healthResource = coreApi.root.addResource('health');
```
guillaumeduboc, line 27: should this ressource be added here or inside the Health function ?

guillaumeduboc: we could use [`resourceForPath`](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-apigateway.IResource.html#resourcewbrforwbrpathpath) method

examples/swarmion-with-cdk/backend/core/package.json
```diff
@@ -36,6 +38,7 @@
     "dotenv": "^16.0.3",
     "esbuild": "^0.15.14",
     "eslint": "^8.27.0",
+    "json-schema-to-ts": "^2.6.0",
```
guillaumeduboc, line 41: you shouldn't need this if you bump the `@swarmion/serverless-contracts` package

fargito:
Closing this as we are rather going to use the CDK generator from #478

PR #353, opened 2022-11-23
TypeScript project references and Nx Cloud

adriencaccia:
- Implement a correct usage of TypeScript project references using this article: https://moonrepo.dev/docs/guides/javascript/typescript-project-refs
- Setup Nx Cloud

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,6 @@ event.json
 
 # Nx caching
 nx-cache
+
+# Build cache manifests
+*.tsbuildinfo
--- a/contracts/orchestrator-contracts/tsconfig.json
+++ b/contracts/orchestrator-contracts/tsconfig.json
@@ -1,10 +1,13 @@
 {
-  "extends": "../../tsconfig.json",
+  "extends": "../../tsconfig.options.json",
   "compilerOptions": {
-    "baseUrl": "src",
-    "composite": true,
-    "outDir": "./dist/types"
+    "baseUrl": "src"
   },
-  "exclude": ["./dist"],
-  "include": ["./**/*.ts"]
+  "references": [
+    {
+      "path": "../../packages/serverless-contracts/tsconfig.build.json"
+    }
+  ],
+  "include": ["./**/*.ts"],
+  "exclude": ["./dist"]
 }
--- a/nx.json
+++ b/nx.json
@@ -2,7 +2,7 @@
   "npmScope": "swarmion",
   "tasksRunnerOptions": {
     "default": {
-      "runner": "nx/tasks-runners/default",
+      "runner": "@nrwl/nx-cloud",
       "options": {
         "cacheableOperations": [
           "build",
@@ -14,7 +14,8 @@
           "test-circular",
           "test-unit"
         ],
-        "cacheDirectory": "nx-cache"
+        "cacheDirectory": "nx-cache",
+        "accessToken": "MzY4MGJjNWEtZmE5YS00ZWQ2LTgzOTEtYzY0Y2MzOTFlMGJkfHJlYWQtd3JpdGU="
       }
     }
   },
--- a/nx.json
+++ b/nx.json
@@ -30,7 +30,7 @@
     "package": {
       "inputs": ["production", "^production"],
       "dependsOn": ["^package"],
-      "outputs": ["{projectRoot}/dist"]
+      "outputs": ["{projectRoot}/dist", "{projectRoot}/tsconfig.tsbuildinfo"]
     },
     "deploy": {
       "inputs": ["production", "^production"],
--- a/packages/create-swarmion-app/package.json
+++ b/packages/create-swarmion-app/package.json
@@ -35,7 +35,7 @@
     "test": "pnpm test-linter && pnpm test-type && pnpm test-unit && pnpm test-circular",
     "test-circular": "pnpm depcruise --validate .dependency-cruiser.js .",
     "test-linter": "pnpm linter-base-config .",
-    "test-type": "tsc --noEmit --emitDeclarationOnly false",
+    "test-type": "tsc",
     "test-unit": "vitest run --coverage --passWithNoTests",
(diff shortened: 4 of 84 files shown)


adriencaccia:
@fargito WDYT?

nx.json
```diff
@@ -14,7 +14,8 @@
           "test-circular",
           "test-unit"
         ],
-        "cacheDirectory": "nx-cache"
+        "cacheDirectory": "nx-cache",
+        "accessToken": "MzY4MGJjNWEtZmE5YS00ZWQ2LTgzOTEtYzY0Y2MzOTFlMGJkfHJlYWQtd3JpdGU="
```
fargito, line 18: @adriencaccia is this a real one?

tsconfig.json
```diff
@@ -1,22 +1,18 @@
...
+    { "path": "./packages/create-swarmion-app/tsconfig.json" },
+    { "path": "./packages/eslint-plugin/tsconfig.build.json" },
+    { "path": "./packages/nx-plugin/tsconfig.build.json" },
+    { "path": "./packages/serverless-cdk-plugin/tsconfig.build.json" },
+    { "path": "./packages/serverless-configuration/tsconfig.build.json" },
+    { "path": "./packages/serverless-contracts/tsconfig.build.json" },
+    { "path": "./packages/serverless-contracts-plugin/tsconfig.build.json" },
+    { "path": "./packages/serverless-helpers/tsconfig.build.json" },
+    { "path": "./user-docs/cloudfront/tsconfig.json" },
+    { "path": "./user-docs/documentation/tsconfig.json" }
```
fargito, line 16: This means we have to update this file when generating a new package/service right?

contracts/orchestrator-contracts/tsconfig.json
```diff
@@ -1,10 +1,13 @@
 {
-  "extends": "../../tsconfig.json",
+  "extends": "../../tsconfig.options.json",
```
fargito, line 2: Shouldn't this change be in `tsconfig.build.json` instead?

fargito (changes requested):
That would be cleaner indeed!

However, I'm confused to whether this change must apply only for `tsconfig.json`, `tsconfig.build.json` or both in libraries.

Also, we need to update generators as well to reflect this change.

.gitignore
```diff
@@ -43,3 +43,6 @@ event.json
 
 # Nx caching
 nx-cache
+
+# Build cache manifests
+*.tsbuildinfo
```
fargito, line 48: I'm also wondering whether this should be cached by Nx as well. In this case, it should be included in the root Nx config

nx.json
adriencaccia: It is yes!

tsconfig.json
adriencaccia: Yes, that way we can run [`tsc --build --verbose` at the root-level](https://moonrepo.dev/docs/guides/javascript/typescript-project-refs#on-all-projects)

.gitignore
adriencaccia: You are completely right 😉

contracts/orchestrator-contracts/tsconfig.json
adriencaccia: I followed [the guide](https://moonrepo.dev/docs/guides/javascript/typescript-project-refs#on-all-projects), and everything is working as expected

adriencaccia:
> That would be cleaner indeed!
> 
> However, I'm confused to whether this change must apply only for `tsconfig.json`, `tsconfig.build.json` or both in libraries.
> 
> Also, we need to update generators as well to reflect this change.

To have a better understanding of the impacts, I would like to test on a bigger repo with known issues (@CorentinDoue I am looking at yours 👀). I will try to migrate it the same way I did in this PR and check if it solves some issues.

tsconfig.json
fargito: This needs to be populated in generators right?

packages/create-swarmion-app/package.json
```diff
@@ -35,7 +35,7 @@
     "test": "pnpm test-linter && pnpm test-type && pnpm test-unit && pnpm test-circular",
     "test-circular": "pnpm depcruise --validate .dependency-cruiser.js .",
     "test-linter": "pnpm linter-base-config .",
-    "test-type": "tsc --noEmit --emitDeclarationOnly false",
+    "test-type": "tsc",
```
guillaumeduboc, line 38: why don't we do this modification in every package ?

contracts/orchestrator-contracts/tsconfig.json
```diff
@@ -1,10 +1,13 @@
...
-    "baseUrl": "src",
-    "composite": true,
-    "outDir": "./dist/types"
+    "baseUrl": "src"
   },
+  "references": [
+    {
+      "path": "../../packages/serverless-contracts/tsconfig.build.json"
+    }
+  ],
```
guillaumeduboc, line 10: is this still needed ? 
it is already specified in the root `tsconfig.options.json`

packages/create-swarmion-app/package.json
adriencaccia: To make `tsc --build` work in both packages and apps, we need to differentiate.
In apps, we will have `"test-type": "tsc",` and `"test-type": "tsc --noEmit --emitDeclarationOnly false",` in packages.

tsconfig.json
adriencaccia: I upgrade them!

adriencaccia:
@fargito It's done 🎉 
I updated the generators and modified the examples.
I didn't have time to test on @CorentinDoue's project.
But I think we can merge anyway, the cleaning of the configs in itself is a good step forward.
And there are other cool things about PR 😉 
WDYT?

fargito (commented):
LGTM, thanks a lot for the hard work! :rocket:

fargito (approved):

PR #357, opened 2022-11-28
Fix TS module error when trying to deploy

### LINKED ISSUE #356: TS module error when trying to deploy
Hello,

On `v0.14.0`, when trying to deploy the whole monorepo, I have the error:
```
    ✖  nx run backend-orchestrator:deploy
       > @swarmion/backend-orchestrator@0.14.0 deploy ........./swarmion/backend/orchestrator
       > serverless deploy
       
       Environment: linux, node 16.17.0, framework 3.25.0 (local), plugin 6.2.2, SDK 4.3.2
       Docs:        docs.serverless.com
       Support:     forum.serverless.com
       Bugs:        github.com/serverless/serverless/issues
       
       Error:
       Cannot load "serverless.ts": Initialization error: /home/fargito/Documents/theodo/swarmion/backend/orchestrator/serverless.ts:1
       import { listDeploymentsContract, onDeploymentRequestedContract, requestSyncDeploymentContract, } from '@swarmion/orchestrator-contracts';
       ^^^^^^
```
I believe the TS `"module"` setting is incorrect for Serverless services. It is set to `esnext` and should be `node16`.
However, I agree that `esnext` should be the value for frontend services.

I think we should set `"module": "node16"` in the root tsconfig and override it only in frontend packages in the examples.

@adriencaccia WDYT?

adriencaccia:
Fix #356

Also adds a release workflow to automatically generate GitHub release notes on releases.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -124,7 +124,7 @@ module.exports = {
       ],
       parser: '@typescript-eslint/parser',
       parserOptions: {
-        project: 'tsconfig.json',
+        project: 'tsconfig.eslint.json',
       },
       settings: { 'import/resolver': { typescript: {} } },
       rules: {
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,21 @@
+name: 🔖 Release
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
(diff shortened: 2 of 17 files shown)


fargito (approved):
Great! Thanks 😄

PR #365, opened 2022-12-09
Feat/add output schemas for non 200 status codes

Sc0ra:


packages/serverless-contracts/src/contracts/apiGateway/__mocks__/httpApiGatewayContract.ts
```diff
@@ -45,5 +58,8 @@ export const httpApiGatewayContractMock = new ApiGatewayContract({
   queryStringParametersSchema,
   headersSchema,
   bodySchema,
-  outputSchema,
+  outputSchemas: {
+    [StatusCodes.OK]: outputSchema,
+    [StatusCodes.BAD_REQUEST]: outputSchema2,
+  },
```
fargito, line 64: Maybe it would be good to have both contracts mocks

packages/serverless-contracts/package.json
```diff
@@ -71,6 +71,7 @@
     "concurrently": "^7.6.0",
     "dependency-cruiser": "^12.1.0",
     "eslint": "^8.28.0",
+    "http-status-codes": "^2.2.0",
```
adriencaccia, line 74: This should be listed in `dependencies`, as it will be used in the exported types.
I think there are other` devDependencies` that should be moved up also, can you check and move them too?

adriencaccia (changes requested):

packages/serverless-contracts/src/contracts/apiGateway/apiGatewayContract.ts
```diff
@@ -122,7 +126,40 @@ export class ApiGatewayContract<
...
+    path: Path;
+    method: Method;
+    integrationType: IntegrationType;
+    pathParametersSchema: PathParametersSchema;
+    queryStringParametersSchema: QueryStringParametersSchema;
+    headersSchema: HeadersSchema;
+    bodySchema: BodySchema;
+    // a object mapping http status codes with JSONSchema used to validate the outputs and infer their types (Same constraints).
+    outputSchemas: OutputSchemas;
+    authorizerType: AuthorizerType;
```
fargito, line 141: Duplicate JSDoc

packages/serverless-contracts/package.json
adriencaccia: Given that, you can also change the commit message to `feat(serverless-contracts): ...` so that it will be in the correct place in the release notes

packages/serverless-contracts/src/contracts/apiGateway/types/lambdaHandler.ts
```diff
@@ -104,14 +105,17 @@ export type InternalSwarmionApiGatewayHandler<
  */
 export type SwarmionApiGatewayHandler<
   Contract extends ApiGatewayContract,
+  ShorthandOutput extends boolean = true,
```
fargito, line 108: Documenter ce type + la propriété shorthand

fargito (changes requested):
Fix the export of types

packages/serverless-contracts/src/contracts/apiGateway/apiGatewayContract.ts
fargito: Or maybe on the implementation props rather than the overloads

packages/serverless-contracts/src/contracts/apiGateway/types/common.ts
```diff
@@ -24,7 +25,19 @@ export type BodyType<Contract extends ApiGatewayContract> =
...
+export type FullOutputType<Contract extends ApiGatewayContract> = O.UnionOf<
+  OutputsType<Contract>
+>;
+
 export type OutputType<Contract extends ApiGatewayContract> =
-  Contract['outputSchema'] extends JSONSchema
-    ? FromSchema<Contract['outputSchema']>
-    : void;
+  | FullOutputType<Contract>
+  | OutputsType<Contract>[200]['body'];
```
fargito, line 43: JSDoc on these types

PR #402, opened 2022-12-15
feat(serverless-helpers): export esbuild config and improve getCdkHandlerPath

fargito:
This PR adds two new features:
- export a default esbuild config from `serverless-helpers`. This can reduce the boilerplate on the Swarmion projects and can enable us to keep it up-to-date for everyone.
- improve the `getCdkHandlerPath` with optional arguments to override `fileName` and `extension`. This is useful in environments where you need to include js handlers directly in your CDK config.

@adriencaccia WDYT?

### DIFF
--- a/packages/serverless-helpers/src/configuration/esbuildConfig.ts
+++ b/packages/serverless-helpers/src/configuration/esbuildConfig.ts
@@ -1,0 +1,33 @@
+import type { BuildOptions } from 'esbuild';
+
+type EsbuildOptions = Omit<BuildOptions, 'watch'>;
+
+/**
+ * The interface for the Swarmion recommended configuration
+ *
+ * See https://github.com/floydspace/serverless-esbuild/blob/master/src/types.ts#L24
+ * And https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda_nodejs.BundlingOptions.html
+ */
+export interface SwarmionEsbuildConfig extends EsbuildOptions {
+  concurrency?: number;
+  packager: PackagerId;
+  exclude: '*' | string[];
+}
+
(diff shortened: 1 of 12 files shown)


packages/serverless-helpers/src/configuration/esbuildConfig.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * the recommended esbuild configuration for Swarmion
+ */
+export const recommendedEsbuildConfig = {
```
adriencaccia, line 4: We should export it with a type.
We can create our own, based on https://github.com/floydspace/serverless-esbuild/blob/master/src/types.ts#L24.

adriencaccia: We can also use some parts of https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_lambda_nodejs.BundlingOptions.html

packages/serverless-helpers/src/helpers/getCdkHandlerPath.ts
```diff
@@ -1,6 +1,25 @@
...
+  fileName?: string;
+};
+
+/**
+ * Helper to be used in CDK config.ts files to retrieve the path of the handler.
+ *
+ * Use it with `__dirname` as the first argument to reference to neighboring files.
+ * You can also customize the filename and extension.
+ */
+export const getCdkHandlerPath = (
```
adriencaccia, line 12: Could we add some unit tests here? 😇

adriencaccia (changes requested):
Those are some nice additions!
It seems that there is a problem with SonarCloud and the coverage report. I'll investigate and try to fix it 😉

PR #417, opened 2022-12-23
feat(serverless-contracts): make `undefined` contract parameters optional

Sc0ra:


packages/serverless-contracts/src/contracts/apiGateway/apiGatewayContract.ts
```diff
@@ -81,7 +80,7 @@ export class ApiGatewayContract<
      *
      * Also please note that for Typescript reasons, you need to explicitly pass `undefined` if you don't want to use the schema.
```
fargito, line 81: There are a couple online comments we need to remove

packages/serverless-contracts/src/contracts/apiGateway/apiGatewayContract.ts
```diff
@@ -155,3 +152,15 @@ export class ApiGatewayContract<
     } as unknown as JSONSchema;
   }
 }
+
+export type GenericApiGatewayContract = ApiGatewayContract<
```
fargito, line 156: Could you add a bit of JSDoc on this?

fargito (changes requested):
This is truly awesome! Thanks a lot @Sc0ra , this will make contacts a lot easier to define! 

A few minor documentation changes but otherwise, this is perfect!

PR #428, opened 2023-01-06
feat(serverless-contracts): allow to type request context more precisely

Sc0ra:


packages/serverless-contracts/src/contracts/apiGateway/__tests__/httpLambdaHandler.test.ts
```diff
@@ -41,7 +41,7 @@ describe('apiGateway lambda handler', () => {
           requestContext,
         }) => {
           await Promise.resolve();
-          const myCustomClaim = requestContext.authorizer.claims.foo ?? '';
+          const myCustomClaim = requestContext.authorizer.claims.foo;
```
fargito, line 44: Could you test with another property? This kind of invalidates the former test, which checks that we get the correct type depending on the authorizer

fargito (approved):

PR #429, opened 2023-01-06
chore: check nextjs starter

fargito:


scripts/check-examples.sh
```diff
@@ -6,7 +6,7 @@ pnpm nx run create-swarmion-app:build
 BASE_DIR=$(pwd)
 TEMP_DIR=$(mktemp -d)
 
-for template in swarmion-starter swarmion-full-stack; do
+for template in swarmion-starter swarmion-full-stack swarmion-with-next; do
```
adriencaccia, line 9: We should loop directly on the `examples` folder, something like

```suggestion
for template in examples/*; do
```

PR #436, opened 2023-01-06
chore(ci): run check example in parallel

Sc0ra:


fargito (changes requested):
Great!

I think you need to change the trigger for the check examples workflow from `check-examples.sh` to `check-example.sh`

fargito (approved):

PR #443, opened 2023-01-09
examples: migrate from useAsync to react-query and remove redux

fargito:
This PR:
- replaces `@react-hookz/web` with `react-query`
- remove Redux from the fullstack example. Check out the diff for the implementation details


examples/swarmion-full-stack/frontend/app/package.json
```diff
@@ -27,12 +27,12 @@
     "@mui/lab": "5.0.0-alpha.114",
```
adriencaccia, line 27: Could you rename the commit to `examples(full-stack): ...` so that it appears the right way in the release motes?

adriencaccia (changes requested):
Nice work, can you change the `with-next` example also?

adriencaccia (approved):
LGTM 👍

PR #447, opened 2023-01-09
chore(docs): deploy preview only when docs are changed

adriencaccia:


.github/workflows/pr-docs-preview.yml
```diff
@@ -0,0 +1,50 @@
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
+      - 'user-dopcs/**'
```
fargito, line 5:
```suggestion
      - 'user-docs/**'
```

PR #457, opened 2023-01-13
Docs : fix `serverless-contracts` examples for `getHandler`

guillaumeduboc:


user-docs/documentation/docs/getting-started/5-add-a-new-serverless-contract-library.mdx
```diff
@@ -191,16 +191,16 @@ Then, you can use the `getHandler` feature to define a type safe handler of your
 import { getHandler } from '@swarmion/serverless-contracts';
 import { getUserContract } from '@my-project-name/users-contracts';
 
-export const handler = getHandler(getUserContract, async event => {
+export const handler = getHandler(getUserContract)(event => {
```
fargito, line 194:
```suggestion
export const handler = getHandler(getUserContract)(async event => {
```

user-docs/documentation/docs/getting-started/5-add-a-new-serverless-contract-library.mdx
```diff
@@ -191,16 +191,16 @@ Then, you can use the `getHandler` feature to define a type safe handler of your
...
   const { userId } = event.pathParameters; // userId will be typed as string
 
   console.log(userId);
 
-  return {
+  return Promise.resolve({
```
fargito, line 199:
```suggestion
  return {
```

user-docs/documentation/docs/getting-started/5-add-a-new-serverless-contract-library.mdx
guillaumeduboc: We are not `await`ing anything in our functions

user-docs/documentation/docs/getting-started/5-add-a-new-serverless-contract-library.mdx
guillaumeduboc: this is a workaround to make our handler asynchronous

fargito: I see, but IMO it gives an incorrect example to new users that might think this is the correct way to define handlers instead of using async functions. I would prefer to keep the async function, remove `Promise.resolve` from the return and add `await Promise.resolve();` inside the handler with a message that explicitely indicates that it must be replaced by the business logic.
WDYT @guillaumeduboc ?

guillaumeduboc: yes I agree, with a comment on the side indicating it stands for the business logic

PR #461, opened 2023-01-13
feat(serverless-contracts): get mock handler event feature

### LINKED ISSUE #376: Build test events generators for Serverless contracts
I was thinking, wouldn't it be great to expose a serverless-contracts feature called `buildTestEvent` to bring even more type-safety to tests?

It could be used like:
```
// handler.test.ts

const { event, context, callback } = buildHandlerTestEvent(myContract)({
    // type-safe values from the contract...
});

const result = await main(event, context, callback);

expect(result.xxxx).toBe('yyyy') // whatever assertion on the result

```

Therefore we would remove the need to directly instantiate `getAPIGatewayV2EventRequestContextMock` or other mock builders. This function would directly populate the event context in the test event.

WDYT?

_Originally posted by @fargito in https://github.com/swarmion/swarmion/pull/295#issuecomment-1334977137_

Sc0ra:
Fix #376

### DIFF
--- a/packages/serverless-contracts/src/contracts/apiGateway/features/mockHandlerInput.ts
+++ b/packages/serverless-contracts/src/contracts/apiGateway/features/mockHandlerInput.ts
@@ -1,0 +1,101 @@
+import { Context } from 'aws-lambda';
+import { JSONSchemaFaker, Schema } from 'json-schema-faker';
+import merge from 'lodash/merge';
+import seedrandom from 'seedrandom';
+
+import {
+  getAPIGatewayEventRequestContextMock,
+  getAPIGatewayV2EventRequestContextMock,
+} from '@swarmion/serverless-helpers';
+
+import { GenericApiGatewayContract } from '../apiGatewayContract';
+import {
+  ApiGatewayEvent,
+  BodyType,
+  CustomRequestContextType,
+  HandlerCallback,
+  HandlerEventType,
+  HeadersType,
+  PathParametersType,
+  QueryStringParametersType,
+} from '../types';
+import {
+  ApiGatewayAuthorizerType,
+  ApiGatewayIntegrationType,
+} from '../types/constants';
+import { getHandlerContextMock } from '../utils/mockHandlerContext';
+
+let seed = 'MySuperSwarmionSeed';
+
+export const setMockHandlerInputSeed = (inputSeed: string): void => {
+  seed = inputSeed;
+};
+
+export const getMockHandlerInput = <
+  Contract extends GenericApiGatewayContract,
+  IntegrationType extends ApiGatewayIntegrationType = Contract['integrationType'],
+  AuthorizerType extends ApiGatewayAuthorizerType = Contract['authorizerType'],
+  AdditionalArgs extends unknown[] = never[],
+  CustomEvent = HandlerEventType<
+    IntegrationType,
+    AuthorizerType,
+    PathParametersType<Contract>,
+    QueryStringParametersType<Contract>,
+    HeadersType<Contract>,
+    CustomRequestContextType<Contract>,
+    BodyType<Contract>
+  >,
+  Event = ApiGatewayEvent<IntegrationType, AuthorizerType>,
+  Callback = HandlerCallback<IntegrationType>,
+>(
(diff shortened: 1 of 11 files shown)


user-docs/documentation/docs/how-to-guides/1-use-serverless-contracts/1-api-gateway.md
```diff
@@ -306,6 +306,57 @@ export const main = applyHttpMiddlewares(handler, {
...
+});
+```
+
+The result value for mockEvent will be something like
+
+```ts
+const mockEvent = {
+  pathParameters: {
+    userId: '15',
+    pageNumber: '42', // Randomly generated
```
GuillaumeLagrange, line 329: This may be more on the `json-schema-faker` side, but is there a way to fix the seed for the random generation ?

I think having non-deterministic tests is not a good practice

packages/serverless-contracts/src/contracts/apiGateway/features/mockHandlerInput.ts
```diff
@@ -0,0 +1,99 @@
...
+  Event = ApiGatewayEvent<IntegrationType, AuthorizerType>,
+  Callback = HandlerCallback<IntegrationType>,
+>(
+  contract: Contract,
+  partialEvent: Partial<CustomEvent> = {},
+  ...additionalArgs: AdditionalArgs
+): [Event, Context, Callback, ...AdditionalArgs] => {
+  const defaultBody =
+    contract.bodySchema !== undefined
+      ? JSONSchemaFaker.generate(contract.bodySchema as Schema)
```
GuillaumeLagrange, line 50: you could write a local `getDefault` function that takes the key of the contract as an argument and return the schema faker generate output, since you basically wrote the same thing 5 times

packages/serverless-contracts/src/contracts/apiGateway/features/mockHandlerInput.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  const event = merge(defaultEvent, partialEvent);
+
+  const stringifiedEvent = {
+    ...event,
+    body: event.body !== undefined ? JSON.stringify(event.body) : undefined,
+  };
+
+  return [
+    stringifiedEvent as Event,
```
GuillaumeLagrange, line 94: naive question ? Do you need the cast here ? Isn't explicitely typing `stringifiedEvent` as `Event` not enough ?

Sc0ra: 'Event' could be instantiated with an arbitrary type which could be unrelated to
:'(

GuillaumeLagrange (approved):

user-docs/documentation/docs/how-to-guides/1-use-serverless-contracts/1-api-gateway.md
```diff
@@ -306,6 +306,96 @@ export const main = applyHttpMiddlewares(handler, {
...
+
+If you need to fix the random generation used by this `getMockHandlerInput`, you can use the `setMockHandlerInputSeed` feature:
+
+```ts
+import {
+  getMockHandlerInput,
+  setMockHandlerInputSeed,
+} from '@swarmion/serverless-contracts';
+
+setMockHandlerInputSeed('42'); // Will always generate the same random values
```
adriencaccia, line 390: How does `faker` work? Is the default behavior to generate different values each time until a seed is used?
If yes, then we can keep it, though adding a small note to their doc would be nice for users without knowledge of `faker`'s API.
If not, let's do the same as them and return the same values every time, and give the users the ability to change it.

adriencaccia (changes requested):

user-docs/documentation/docs/how-to-guides/1-use-serverless-contracts/1-api-gateway.md
Sc0ra: faker default behavior is random data unitl a seed is used.

Where would you like to add documentation ?

adriencaccia (approved):

PR #462, opened 2023-01-13
examples(swarmion-full-stack): add exported files to shared library

guillaumeduboc:
Handled file types :

- [x] svg
- [x] png
- [x] jpg 
- [x] webp

### DIFF
--- a/examples/swarmion-full-stack/frontend/shared/src/components/NotFound/__tests__/NotFound.test.tsx
+++ b/examples/swarmion-full-stack/frontend/shared/src/components/NotFound/__tests__/NotFound.test.tsx
@@ -2,13 +2,6 @@ import { render, screen } from '@testing-library/react';
 
 import NotFound from '../NotFound';
 
-vi.mock('assets', () => ({
-  NotFoundIcon: 'svg',
-  notFoundJpgUrl: 'jpg',
-  notFoundPngUrl: 'png',
-  notFoundWebpUrl: 'webp',
-}));
-
 describe('NotFound component', () => {
   it("should display the heading 'Page not found'", () => {
     render(<NotFound />);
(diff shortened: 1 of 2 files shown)


examples/swarmion-full-stack/frontend/shared/src/assets/icons/assetAsString.d.ts
```diff
@@ -0,0 +1,12 @@
+declare module '*.png' {
```
fargito, line 1: Hmm, is this the place for this declaration?

adriencaccia:
I think we should have an example for exposing things like `png`, `jpg` or `webp` files.
Here with the `svgr` plugin, all `.svg` files are directly loaded into the `js` code.

examples/swarmion-full-stack/frontend/shared/src/assets/icons/assetAsString.d.ts
guillaumeduboc: @fargito where would you rather put it ?

guillaumeduboc:
Latest screenshot : 2023-01-13

Capture d’écran 2023-01-13 à 18 17 47

examples/swarmion-full-stack/frontend/shared/src/components/NotFound/__tests__/NotFound.test.tsx
```diff
@@ -2,6 +2,13 @@ import { render, screen } from '@testing-library/react';
 
 import NotFound from '../NotFound';
 
+vi.mock('assets', () => ({
+  NotFoundIcon: 'svg',
+  notFoundJpgUrl: 'jpg',
+  notFoundPngUrl: 'png',
+  notFoundWebpUrl: 'webp',
+}));
```
adriencaccia, line 10: Let's go and try to find a way to always mock `*.{svg,jpg,png,webp}` files. It should be done automatically instead of doing it in every test files.

adriencaccia (changes requested):

adriencaccia (approved):
This is good enough for now, thanks @guillaumeduboc 🎉

PR #466, opened 2023-01-16
feat: use nx subtargets for test commands

### LINKED ISSUE #448: Make the root `test` target use sub targets' nx cache
Currently, if we run

`pnpm test-linter` at the root of the workspace, then run `pnpm test`, the `test` target does not make use of the `test-linter` nx cache.

We could remove the `test` target in workspaces, and keep the `test` target in root, which would only call 
```
 nx run-many --targets=test-linter,test-type,test-circular,test-unit --all
```
[relevant nx example (last one)](https://nx.dev/packages/nx/documents/run-many#examples)

What do you think ?

fargito:
Fixes #448 

- use the test subcommands in order to improve nx cache for tests
- align all `nx.json` configurations


examples/swarmion-full-stack/nx.json
```diff
@@ -5,9 +5,9 @@
       "runner": "nx/tasks-runners/default",
       "options": {
         "cacheableOperations": [
+          "bootstrap",
```
guillaumeduboc, line 8: can nx cache a script that is not a target ?

adriencaccia (approved):

examples/swarmion-full-stack/nx.json
fargito: Yes, it will simply not run it

PR #478, opened 2023-01-18
feat(nx-plugin): add CDK service generator

fargito:


packages/nx-plugin/schemas/cdk-service/files/resources/index.ts__tmpl__
```diff
@@ -0,0 +1,24 @@
...
+import { App } from 'aws-cdk-lib';
+import dotenv from 'dotenv';
+
+import {
+  defaultEnvironment,
+  region,
+  sharedParams,
+} from '@<%= workspaceName %>/serverless-configuration';
+
+import { <%= capitalize(pascalName) %>Stack } from './stack';
```
adriencaccia, line 12: `pascalName` should already have the first letter capitalized since it comes from `className`:
image

adriencaccia: I tested it, and it works!
So you can remove the `capitalize` function and its uses

adriencaccia (changes requested):
Huge work 💪 

I have a small feedback 😉

packages/nx-plugin/schemas/cdk-service/files/resources/index.ts__tmpl__
fargito: Indeed! I removed this

adriencaccia (approved):

PR #480, opened 2023-01-20
chore(bench): add codspeed bench

adriencaccia:
Add [CodSpeed](https://codspeed.io/) continuous benchmarking tool.

In order to make it work, the `serverless-contracts` package has been switched to `esm`. This is something that needed to be done at some time, so it is a nice opportunity to start experimenting with it 😄


adriencaccia:
@fargito WDYT?

Especially the part about `esm`. I tested it on my end and found it's working. I would like your opinion/tests before merging this 😉

packages/serverless-contracts/package.json
```diff
@@ -32,7 +34,7 @@
     "package-types": "tsc -p tsconfig.build.json",
     "package-types-aliases": "tsc-alias -p tsconfig.build.json",
     "test": "nx run test-linter && nx run test-type && nx run test-unit && nx run test-circular",
-    "test-circular": "pnpm depcruise --validate .dependency-cruiser.js .",
+    "test-circular": "pnpm depcruise --config -- .",
```
fargito, line 37: Why was it necessary to remove the `--validate` option?

packages/serverless-contracts/package.json
```diff
@@ -19,10 +19,12 @@
...
     "dist"
   ],
-  "main": "dist/index.cjs.js",
-  "module": "dist/index.esm.js",
+  "type": "module",
+  "main": "dist/index.cjs",
+  "module": "dist/index.js",
   "types": "dist/types/index.d.ts",
   "scripts": {
+    "bench": "pnpm tsx src/__benches__/index.ts",
```
fargito, line 27: Maybe this folder should be outside of `src`?
We need to make sure this is not included in the code shipped to npm

fargito (changes requested):
Very nice!
Let's ensure we don't ship the benches to npm though

packages/serverless-contracts/package.json
adriencaccia: It is explained here: https://github.com/sverweij/dependency-cruiser/blob/develop/doc/cli.md#--config---validate

Basically, `--validate` and `--config` are aliases, and we need to use it to have the default config (in our case `.dependency-cruiser.cjs`) loaded

packages/serverless-contracts/package.json
adriencaccia: Since nothing from this folder is exported from `src/index.ts`, there will not be any `js` code exported. And for types the `src/__benches__` folder is excluded from the `tsconfig.build.json`.

I wanted to keep the benches in the `src` folder because, in the long run, our benches will be done through `vitest bench` (once CodSpeed supports it), and our benches will live right next to the source code as tests do.

WDYT?

packages/serverless-contracts/package.json
fargito: Perfect!

packages/serverless-contracts/package.json
fargito: I see, and in this case `src/__benches__/index.ts` will disappear right ? In order to minimize the rework when that is done, we can put it in `__benches__` and directly import from all the benches subdirectories (skip all the `index.ts` exports). WDYT?

adriencaccia: Yes you are completely right, I will do that 😉

fargito (approved):

PR #483, opened 2023-01-23
ci(cd): clean & fix examples

fargito:


examples/swarmion-with-next/.github/workflows/merge-main.yml
```diff
@@ -3,10 +3,17 @@ on:
     branches:
       - main
 
-concurrency: deploy_main
-
```
adriencaccia, line 7: Why remove this?
This prevents concurrent cloudformation deployments, I think we should keep it.

Maybe add a comment if it is not explicit enough

examples/swarmion-with-next/.github/workflows/merge-main.yml
```diff
@@ -26,43 +31,33 @@ jobs:
...
+      - name: '🧪 Run tests'
+        run: pnpm nx affected --targets=test-linter,test-type,test-unit,test-circular --parallel=2
       - name: Configure AWS Credentials
-        uses: aws-actions/configure-aws-credentials@master
+        uses: aws-actions/configure-aws-credentials@v1-node16
         with:
-          aws-region: ${{ steps.get-stack-region.outputs.AWS_REGION }}
-          role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }} # Create the role in AWS and add it to the secrets
+          aws-region: ${{ env.AWS_REGION }}
+          role-to-assume: ${{ secrets.AWS_ROLE_ARN_PRODUCTION }}
```
adriencaccia, line 61: We should change the docs to document that these should be defined/changed

adriencaccia (changes requested):

examples/swarmion-with-next/.github/workflows/merge-main.yml
fargito: This is a good catch, it had been removed from the other examples and I copy-pasted 😆

examples/swarmion-with-next/.github/workflows/merge-main.yml
fargito: Done

user-docs/documentation/docs/how-to-guides/4-configure-ci-cd.md
```diff
@@ -6,17 +6,35 @@ sidebar_position: 4
...
+We strongly recommend the second option, as it will remove the risk of losing long-lived credentials. The Swarmion CI/CD from all starters, uses this method by default.
 
-This will remove the need to store long lived credentials, and thus the need to specify a policy (we can grant admin access). If you do so, you can skip the following sections.
+### With OpenID Connect (OIDC)
+
+If you use Github Actions, we can recommend to follow [these guidelines](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services) to setup OIDC for your pipeline.
+
+If you are using the Swarmion generated CI, once you have created the role, retrieve its ARN and place it in a Github Actions secret named `AWS_ROLE_ARN_PRODUCTION`.
+
+### With an IAM user
```
adriencaccia, line 26: We could separate the two parts with [tabs](https://docusaurus.io/docs/markdown-features/tabs), so it will be clearer that the restriction is not recommend for OIDC

adriencaccia (changes requested):

PR #490, opened 2023-01-27
perf(serverless-contracts): move ajv schema compilation outside of handler invocation

adriencaccia:
This change allows to compile the schema only once, instead of compiling it on every invocation.
On handlers with basic schemas, this change reduces the execution by around 5ms.
With larger schemas, the difference is even more significant.

### DIFF
--- a/packages/serverless-contracts/src/contracts/apiGateway/features/lambdaHandler.ts
+++ b/packages/serverless-contracts/src/contracts/apiGateway/features/lambdaHandler.ts
@@ -52,63 +52,68 @@ export const getApiGatewayHandler =
     AuthorizerType,
     Output,
     AdditionalArgs
-  > =>
-  async (event, context, callback, ...additionalArgs) => {
-    try {
-      const ajv = new Ajv();
+  > => {
+    const ajv = new Ajv();
+    const inputValidator = ajv.compile(contract.inputSchema);
 
-      const parsedEvent = proxyEventToHandlerEvent<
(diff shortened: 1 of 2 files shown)


packages/serverless-contracts/src/contracts/apiGateway/features/lambdaHandler.ts
```diff
@@ -52,63 +52,64 @@ export const getApiGatewayHandler =
...
+        );
 
-      if (contract.outputSchema !== undefined) {
-        const outputValidator = ajv.compile(contract.outputSchema);
-        if (!outputValidator(handlerResponse)) {
-          console.error('Error: Invalid output');
-          console.error(outputValidator.errors);
-          throw createHttpError(400, 'Invalid output');
+        if (contract.outputSchema !== undefined) {
+          const outputValidator = ajv.compile(contract.outputSchema);
```
fargito, line 85: You can move this out of the handler declaration as well right?

fargito (changes requested):
Awesome!

Great catch thanks to Codspeed, congrats @art049 for this

packages/serverless-contracts/src/contracts/apiGateway/features/lambdaHandler.ts
adriencaccia: Nicely spotted 👀

fargito (approved):
🚀

PR #495, opened 2023-01-27
Fix swarmion contracts' dependency

guillaumeduboc:


packages/serverless-contracts/package.json
```diff
@@ -59,6 +58,7 @@
     "@middy/core": "^3.6.2",
     "@middy/error-logger": "^3.6.2",
     "@middy/http-cors": "^3.6.2",
+    "@swarmion/serverless-helpers": "^0.21.3",
```
fargito, line 61: Are you sure about this? What do we use it for?

guillaumeduboc: it is used in the tests and benches

packages/serverless-contracts-plugin/package.json
```diff
@@ -68,7 +67,7 @@
     "vitest": "^0.28.0"
   },
   "peerDependencies": {
-    "@swarmion/serverless-contracts": "^0.16.1",
+    "@swarmion/serverless-contracts": "workspace:^",
```
adriencaccia, line 70: I fear that this might not be portable for repositories not using workspaces, and moreover if they use `yarn` or `npm` instead of `pnpm`.

As we always deploy the two packages together, it makes sense to make sure the two dependencies have the same version. SO let's keep it that way.

Why moving it as a `peerDependency`?

adriencaccia (changes requested):

packages/serverless-contracts-plugin/package.json
guillaumeduboc: I believe `workspace:^` is resolved when publishing to npm and is not in the published `package.json` so there shouldn't be an issue
I used it because it seems the version of serverless-contracts in peerDependencies was not updated

It was already as peerDependency. I think we should leave it there as it should not be owned by the plugin

fargito: Well maybe we should remove the plugin completely, we don't use it and it only makes sense with the Serverless Framework

adriencaccia: > I believe workspace:^ is resolved when publishing to npm and is not in the published package.json so there shouldn't be an issue

Do you have a link that supports this? I cannot find anything on this.

> Well maybe we should remove the plugin completely, we don't use it and it only makes sense with the Serverless Framework

Interesting take, I kind of forgot this could be possible 😅. Maybe we can do this in an other PR.

guillaumeduboc: > Do you have a link that supports this? I cannot find anything on this.

you can find it right here : https://pnpm.io/workspaces#publishing-workspace-packages

adriencaccia (approved):

PR #496, opened 2023-01-27
feat(create-swarmion-app): add swarmion-bare

adriencaccia:
- chore(swarmion-bare): copy swarmion-starter to swarmion-bare
- chore(swarmion-bare): remove core service and contracts
- feat(create-swarmion-app): explicit backend type in starter
- feat(create-swarmion-app): add swarmion-bare

Adresses #494


fargito (changes requested):
I think you need to edit the list of templates in
https://github.com/swarmion/swarmion/blob/main/packages/create-swarmion-app/src/getProjectTemplate.ts#L4-L8

fargito:
I'm not sure why the examples check for `swarmion-starter` and `swarmion-with-next` fail, I'll try and investigate

adriencaccia:
> I'm not sure why the examples check for `swarmion-starter` and `swarmion-with-next` fail, I'll try and investigate

I found the problem and fixed it.

**TL:DR**
`@swarmion/serverless-helpers` had the `@serverless/typescript` package listed as a dependency with version `^3.25.0` and the `swarmion-starter` example had the version `3.27.0` installed because of the `^3.25.0` in the `core` service.
So there was mismatch of the `AWS` type used in the `services/core/serverless.ts` file.

To fix it I changed the `@serverless/typescript` dependency to a peer dependency in the `@swarmion/serverless-helpers` package. I also did it in the `@swarmion/serverless-contracts` package since it used the dependency aswell.

WDYT?

fargito (approved):

PR #498, opened 2023-01-28
feat(nx-plugin): add qualifier to cdk generator

### LINKED ISSUE #491: Enable parallel CDK stacks deployment
When using generated CDK services, it is currently impossible to deploy multiple services in parallel.

This is because each service needs the bootstrap stack, and there is only one per environment.

We should leverage the `qualifier` argument of the CDK to enable this. See https://docs.aws.amazon.com/cdk/v2/guide/bootstrapping.html#bootstrapping-custom-synth

fargito:
fixes #491 

The idea of this fix is:
- each stack gets its bootstrap stack named `CDKToolkit{projectName}`
- each bootstrap stack uses a `qualifier` to ensure resources won't clash
- this `qualifier` must be explicitely passed to `cdk bootstrap`, otherwise it is not applied
- this `qualifier` must be less than 10 chars, so a SHA1 hash is enough

See https://docs.aws.amazon.com/cdk/v2/guide/bootstrapping.html

### DIFF
--- a/packages/nx-plugin/schemas/cdk-service/files/cdk.json__tmpl__
+++ b/packages/nx-plugin/schemas/cdk-service/files/cdk.json__tmpl__
@@ -2,7 +2,9 @@
   "app": "npx ts-node --prefer-ts-exts resources/index.ts",
   "requireApproval": "never",
   "watch": {
-    "include": ["**"],
+    "include": [
+      "**"
+    ],
     "exclude": [
       "coverage",
       "project.json",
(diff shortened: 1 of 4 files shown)


fargito:
Ok, I checked on my project. Once everything is deployed, the old bootstrap stack will still be present, but can safely manually deleted

fargito:
@adriencaccia is that ok for you?

adriencaccia (approved):

Issue #329, opened 2022-10-28
[`EventBridgeContract`] add a `getCdkTrigger` function


fargito:
The current implementation of `getTrigger` is for the Serverless Framework. However, we may want to integrate this solution with the CDK.

Maybe we could also implement it for `ApiGatewayContract` 🤔

fargito:
Having a `pattern` attribute on our contract would be enough for this I think

PR #524, opened 2023-02-10
feat(serverless-configuration): build patterns for eventBridge contracts

fargito:
Fix: #329


packages/serverless-contracts/package.json
```diff
@@ -46,6 +46,7 @@
     "@types/aws-lambda": "^8.10.109",
     "@types/http-errors": "^2.0.1",
     "ajv": "^8.11.2",
+    "aws-cdk-lib": "^2.64.0",
```
adriencaccia, line 49: Having the dependency here will make everyone download the lib, even if they do not use it.

It will also mean that consumers of the package, which can be frontend applications, will install this dependency.

We might need to rethink how we structure our packages, because making the producers and the consumers install dependencies that they will not use is really problematic.

In the meantime, we maybe can add it as an [`optionaDependencies`](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#optionaldependencies)?

adriencaccia (changes requested):

packages/serverless-contracts/package.json
fargito: That is correct, and since it is only for one type, I'll simply copy it in our repo

adriencaccia (approved):

PR #541, opened 2023-02-14
fix/eslint project

adriencaccia:
- chore(eslint): load the right eslint config per workspace
- fix(nx-plugin): load the right eslint config per workspace in VS Code
- examples: load the right eslint config per workspace in VS Code


fargito (approved):
Awesome!

guillaumeduboc (approved):
lgtm

PR #572, opened 2023-03-06
fix(bare template): remove unnecessary workspaces

adelego:
Hey I add an error with ESLint import rule when I created a project from `bare`. It couldn't find a contracts directory because it wasn't created.
This fixed for me


fargito (approved):
Thanks!

fargito:
@allcontributors please add @adelego for code

PR #578, opened 2023-03-10
feat(integration-tests): create package

### LINKED ISSUE #564: Add `@swarmion/integration-tests` with `TestEnvVar` & `syncTestEnvVars`
## Part 1 of https://github.com/swarmion/swarmion/issues/525

Concretely I want to open source the code I did with @MaximeVivier to help manage the integration test environment variables. 
The interface will be something like

```ts 
// cdk stack
stack.node.setContext(SCOPE_CONTEXT_NAME, 'myScope');
// ...
// latter in a construct
new TestEnvVar(this, "API_URL", { value: httpApi.url }); // => create a SSM parameter /testEnvVar/myScope/API_URL
```

```ts
// setup file of the test runner 
await syncTestEnvVars({ scope: 'myScope' }); // get /testEnvVar/myScope/API_URL from SSM or cache and push it into process.env.API_URL
```

```ts
// integration test file

describe("health check", () => {
  it("returns 200", async () => {
    const response = await axios.get(`${TestEnvVars.API_URL}/health`);
    expect(response.status).toBe(200);
  });
});

```

The env vars can be scoped thank a string parameter to avoid importing all of them in the integration tests

`TestEnvVar` also creates & syncs a ts file to get the declared variable types. This file can be committed.

### LINKED ISSUE #585: `syncEnvVarType` is broken
When I use `@swarmion/integration-tests` on Canopia the `syncEnvVarType` util is broken:

```
yarn sls deploy

Error while trying to sync config types: [TypeError] (0 , import_traverse.default) is not a function
```
This is the problem of compatibility between @babel/traverse and ESM: https://github.com/babel/babel/issues/13855

I suppose, it worked on Canopia before because the package was bundled with the old method (babel + cjs first). 

Now `@swarmion/integration-tests` is bundled with tsup and is ESM first 

=> I suppose node load the package as ESM when I run `sls deploy`

Versions:

- Node: `14.18.1`
- Sls framework: `3.25.0`
- `@swarmion/integration-tests`: `0.26.0-alpha.0`

CorentinDoue:
fix #564
fix #585


packages/integration-tests/src/syncTestEnvVarType/syncTestEnvVarType.ts
```diff
@@ -0,0 +1,111 @@
...
+  tsTypeAliasDeclaration,
+  tsTypeAnnotation,
+  tsTypeLiteral,
+} from '@babel/types';
+import fs from 'fs';
+import { parse, print } from 'recast';
+import parser from 'recast/parsers/typescript';
+
+// @ts-expect-error - issue with esm and traverse: https://github.com/babel/babel/issues/13855
+const traverse = _traverse.default as typeof _traverse;
```
CorentinDoue, line 22: Does it still work with cjs ? Because when I did this on Canopia, it worked with locally on vitest but broke the cjs

adriencaccia: As explained in https://github.com/swarmion/swarmion/issues/585#issuecomment-1470895941, the `cjs` version is used when running `cdk ...` commands

adriencaccia:
@CorentinDoue When the file type is created, or a new property is added to `TestEnvVarsType`, each property line is missing a trailing semi-colon `;` character.

For example, the created file with the env variables `API_URL` and `TABLE_NAME` will be:

```typescript
export type TestEnvVarsType = {
  API_URL: string
  TABLE_NAME: string
};
```

You can see that the semi-colons are missing.

adriencaccia:
> @adriencaccia I know. But I can't add it with recast. It's not a part of the typescript syntax but a style preference. So it's not a node in the ast. Somebody can have a prettier config that don't add trailing `;`.
> 
> I don't think it's the role of the lib to handle style preferences.
> 
> WDYT ?

Oh my bad, I did not know that TS interfaces worked without the semi-colons 😲 
It's fine then yeah

adriencaccia (changes requested):
Nice work @CorentinDoue 🎉

adriencaccia (approved):

Issue #582, opened 2023-03-10
Can't deploy orchestrator stack


CorentinDoue:
To reproduce:
1. `git clone`
2. `pnpm i`
3. configure aws profile `swarmion-developer`
4. run `pnpm run deploy` in services/orchestrator
=>
```
✖ Stack swarmion-orchestrator-dev failed to deploy (49s)
Environment: darwin, node 16.19.1, framework 3.27.0 (local), plugin 6.2.3, SDK 4.3.2
Credentials: Local, "swarmion-developer" profile
Docs:        docs.serverless.com
Support:     forum.serverless.com
Bugs:        github.com/serverless/serverless/issues

Error:
CREATE_FAILED: HttpApi (AWS::ApiGatewayV2::Api)
Resource handler returned message: "Invalid format for origin localhost:3000 (Service: AmazonApiGatewayV2; Status Code: 400; Error Code: BadRequestException; Request ID: 922af6bd-3677-4291-b8ff-ae723f3c18c1; Proxy: null)" (RequestToken: 1346b43e-9d2f-ab22-6e95-cd9c924d07da, HandlerErrorCode: GeneralServiceException)
``` 
![image](https://user-images.githubusercontent.com/31917261/224375374-a11ea8c3-460f-4563-8e4d-8afa837efd40.png)

CorentinDoue:
It comes from `apiGatewayCorsAllowedOrigins` from `serverless-configuration`. The prefix `http://`is missing

PR #583, opened 2023-03-10
chore(serverless-configuration): fix apiGatewayCorsAllowedOrigins

CorentinDoue:
`http://` is mandatory in CORS to deploy httpApi

fix https://github.com/swarmion/swarmion/issues/582


adriencaccia (commented):
Thanks @CorentinDoue for the fix!

I will change the commit to be `chore` instead of `fix`. since this change only applies to a private package, we do not want this commit to appear in the release notes

adriencaccia (approved):

PR #586, opened 2023-03-14
Chore : update stylelint config

guillaumeduboc:
Since the update of stylelint to v15, many rules have been deprecated and the config should be updated


adriencaccia (changes requested):
Nice one @guillaumeduboc, love the fact that it simplifies the config and number of `stylelint` dependencies

PR #598, opened 2023-03-24
Chore : update `stylelint` extension config

guillaumeduboc:


packages/nx-plugin/src/generators/library/index.ts
```diff
@@ -1,12 +1,7 @@
 import { formatFiles, Tree, writeJson } from '@nrwl/devkit';
```
adriencaccia, line 1: You can remove the related test cases here: https://github.com/swarmion/swarmion/blob/4c52427f180083796344bc1ac4f8ce582c4e968f/packages/nx-plugin/src/generators/library/library.test.ts#L51

adriencaccia (changes requested):
Noooiiice 👏

packages/nx-plugin/src/generators/library/index.ts
guillaumeduboc: How come the test did not fail ? 👀

adriencaccia (approved):

Issue #615, opened 2023-04-07
@swarmion/nx-plugin:cdk-service creates broken cdk bootstraps 


CorentinDoue:
### Describe the bug

using `nx generate @swarmion/nx-plugin:cdk-service nft-dummy`, the generated service contains some scripts to bootstrap the CDK

```json
{
 "scripts": {
    "bootstrap": "cdk bootstrap --profile swarmion-developer --qualifier s_8swf03cp",
    "bootstrap-production": "cdk bootstrap --context stage=production --qualifier s_8swf03cp",
    "bootstrap-staging": "cdk bootstrap --context stage=staging --qualifier s_8swf03cp",
  }
}
```

the qualifier is built from a hash of the project name 
https://github.com/swarmion/swarmion/blob/27cdea8249c2c122b2ff078260c7ba92c702ab78/packages/nx-plugin/src/generators/cdk-service/typed-json-config/package.json.ts#L9
https://github.com/swarmion/swarmion/blob/7b8edcb1b713b12664b602ab18b35296ce1f9dd1/packages/nx-plugin/src/generators/helpers/normalizeOptions.ts#L35-L39
This hash can contain characters that are not supported as qualifier such `_`

=> The bootstrap fails 💥

```
❯ pnpn run bootstrap

> @swarmion/nft-dummy@1.0.0 bootstrap /Users/cdoue/www-dev/projects/swarmion/services/nft-dummy
> cdk bootstrap --profile swarmion-developer --qualifier s_8swf03cp

 ⏳  Bootstrapping environment aws://032062562436/eu-west-1...
Trusted accounts for deployment: (none)
Trusted accounts for lookup: (none)
Using default execution policy of 'arn:aws:iam::aws:policy/AdministratorAccess'. Pass '--cloudformation-execution-policies' to customize.
CDKToolkitNftDummy: creating CloudFormation changeset...
 ❌  Environment aws://032062562436/eu-west-1 failed bootstrapping: Error [ValidationError]: Template format error: Output FileAssetKeyArn is malformed. The Name field of every Export member must be specified and consist only of alphanumeric characters, colons, or hyphens.
    at Request.extractError (/Users/cdoue/www-dev/projects/swarmion/node_modules/.pnpm/aws-cdk@2.73.0/node_modules/aws-cdk/lib/index.js:275:46127)
[...]

PR #616, opened 2023-04-07
fix(nx-plugin): create hashedProjectName without special characters

CorentinDoue:
fix https://github.com/swarmion/swarmion/issues/615

use `hex` encoding to create a `hashedProjectName` that matches `/^[a-z0-9]{10}$/` and avoid side effects caused by not supported special characters


adriencaccia (approved):

adriencaccia:
@CorentinDoue Rebasing the PR should do the trick and make the CI pass

PR #617, opened 2023-04-07
Feat/remove eslint prettier

Sc0ra:
* Remove eslint-prettier deps
* Directly use prettier for formatting purpose
* Split formatting and linting
* Upgrade @typescript-eslint


.github/workflows/pr.yml
```diff
@@ -33,6 +33,8 @@ jobs:
       - uses: nrwl/nx-set-shas@v3
       - name: Install & cache node dependencies
         uses: ./.github/actions/install-node-deps
+      - name: '💄 Check format'
+        run: pnpm nx affected --targets=format-check --parallel=2
```
adriencaccia, line 37: Let's run it against the whole repo here, it should not take that much time.
There is no need to leverage Nx here

.github/workflows/push-main.yml
```diff
@@ -24,6 +24,8 @@ jobs:
       - uses: nrwl/nx-set-shas@v3
       - name: Install & cache node dependencies
         uses: ./.github/actions/install-node-deps
+      - name: '💄 Check format'
+        run: pnpm nx affected --targets=format-check --parallel=2
```
adriencaccia, line 28: Idem

adriencaccia (changes requested):
Let's do the changes on the examples and generators too

adriencaccia:
@Sc0ra Rebasing the PR should do the trick and make the CI pass

Could you also do the following?

> Let's squash all the commits into two different, one for all the configuration and one for actually running `prettier`.
> 
> Let's do the changes to the examples and generators too. One commit for each.

adriencaccia (changes requested):
Hey, great work @Sc0ra 💪 

Can you change the following commits?
image

The `fix` and `feat` commit types are reserved for commits affecting deployed npm packages, such as `serverless-contracts`. Whenever we are modifying the repo as a whole without modifying the packages, the `chore` type should be used. And when modifying the examples, the `examples` commit type should be used.

Issue #464, opened 2023-01-14
`getFetchRequest` example doesn't resolve the correct URL


guillaumeduboc:
### Issue
`getFetchRequest` is not resolving the proper url

### Reproduction

Create a contract following the tutorial and use a `getFetchRequest`.
```typesctipt
const getUserContract = new ApiGatewayContract({
  id: 'users-getUser',
  path: '/users/{userId}',
  method: 'GET',
  integrationType: 'httpApi',
  pathParametersSchema,
  outputSchema: userEntitySchema,
});
```

use `https://toto.com/tata` or `https://toto.com/tata/` as a `baseUrl`

The `getFetchRequest` will try to fetch the url `https://toto.com/users/{userId}`

guillaumeduboc:
### Reason

This is due to the behavior of the URL module. 

```typescript

new URL('/users/{userId}', 'https://toto.com/tata')   // will return https://toto.com/users/{userId}

new URL('/users/{userId}', 'https://toto.com/tata/')  // will return https://toto.com/users/{userId}

new URL('users/{userId}', 'https://toto.com/tata')   // will return https://toto.com/users/{userId}


new URL('users/{userId}', 'https://toto.com/tata/')   // will return https://toto.com/tata/users/{userId}
```


We don't have this issue with `getAxiosRequest` because axios use their own rule to [combine the url](https://github.com/axios/axios/blob/d83316db4a242252db3a2ed7728cb43f8f8f4189/dist/axios.js#L1872-L1875)


### Quick fix

When using the contract, remove the leading slash in path and add a trailing slash on baseUrl

### Long term fix

use the same combine url as axios to concatenate baseUrl and path before transforming it into a URL object
WDYT ?

adriencaccia:
Hey @guillaumeduboc, I agree with the long-term fix you proposed. We should have unit tests that assure that the URL is resolved correctly.

PRs are welcome!

PR #626, opened 2023-04-16
fix(serverless-contracts): combine baseUrl and path correctly in fetchRequests

PChol22:
this is a fix proposal for issue #464, based on the url combining used by axios

### DIFF
--- a/packages/serverless-contracts/src/contracts/apiGateway/utils/__tests__/combineUrls.test.ts
+++ b/packages/serverless-contracts/src/contracts/apiGateway/utils/__tests__/combineUrls.test.ts
@@ -1,0 +1,30 @@
+import { combineUrls } from '../combineUrls';
+
+const expectedUrl = new URL('https://example.com/base/path');
+
+describe('combineUrls', () => {
+  it.each([
+    {
+      baseUrl: 'https://example.com/base',
+      path: '/path',
+    },
+    {
+      baseUrl: 'https://example.com/base',
+      path: 'path',
+    },
+    {
+      baseUrl: 'https://example.com/base/',
+      path: '/path',
+    },
+    {
+      baseUrl: 'https://example.com/base/',
+      path: 'path',
+    },
+  ])(
+    'should combine $baseUrl and $path into $expected',
+    ({ baseUrl, path }) => {
+      expect(combineUrls(path, baseUrl)).toEqual(expectedUrl);
+      expect(combineUrls(path, new URL(baseUrl))).toEqual(expectedUrl);
+    },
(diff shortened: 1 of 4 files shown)


packages/serverless-contracts/src/contracts/apiGateway/utils/combineUrls.ts
```diff
@@ -0,0 +1,10 @@
+/// <reference lib="dom" />
```
adriencaccia, line 1: Is it really necessary to add this?
I am afraid that it would define global dom types for all the users of the package, even if they use the package only in the backend.

If it is mandatory, we'll have to think of a workaround.

adriencaccia (changes requested):
Thanks @PChol22 for the PR, really glad to see you contributing 🤩 

In order to have better automated release notes, can you change the commit message to something like:
`fix(serverless-contracts): ...`

The idea is to have the name of the affected published package in the scope of the commit.

packages/serverless-contracts/src/contracts/apiGateway/utils/__tests__/combineUrls.test.ts
```diff
@@ -0,0 +1,39 @@
+import { combineUrls } from '../combineUrls';
+
+describe('combineUrls', () => {
+  it('should combine two string urls', () => {
```
adriencaccia, line 4: Use `it.each` to have a single assertion per test case.

packages/serverless-contracts/src/contracts/apiGateway/utils/__tests__/combineUrls.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    expect(combineUrls('path', 'https://example.com/base')).toEqual(
+      new URL('https://example.com/base/path'),
+    );
+
+    expect(combineUrls('path', 'https://example.com/base/')).toEqual(
+      new URL('https://example.com/base/path'),
+    );
+  });
+
+  it('should combine a URL url and a string url', () => {
```
adriencaccia, line 22: Idem

adriencaccia (changes requested):

packages/serverless-contracts/src/contracts/apiGateway/utils/combineUrls.ts
Sc0ra: This is usefull when using fetch, it add type def without any package
In this case, I don't think it is useful so you can remove it

adriencaccia (approved):

Issue #629, opened 2023-04-20
Add option to return validation error instead of "Invalid input" when the body doesn't match the bodySchema of the ApiGatewayContract


clementmarcilhacy:
### Clear and concise description of the problem

As a developer using Swarmion I want see the validation errors like in cloudwatch when I try to access an API with a body that doesn't match the bodyScham instead of a raw string "Invalid input" so that the consumers of my API know what is wrong.

### Suggested solution

add a flag named `returnValidationErrors` and return validation errors if this flag is activated instead of "Invalid input"

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/swarmion/swarmion/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://www.swarmion.dev/docs/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

adriencaccia:
Fixed in #630

PR #630, opened 2023-04-20
feat(serverless-contracts): add option to return validation errors

clementmarcilhacy:


packages/serverless-contracts/src/contracts/apiGateway/__tests__/httpLambdaHandler.test.ts
```diff
@@ -194,6 +194,71 @@ describe('apiGateway lambda handler', () => {
...
+              schemaPath: '#/properties/body/required',
+              keyword: 'required',
+              params: { missingProperty: 'foo' },
+              message: "must have required property 'foo'",
+            },
+          ],
+        }),
+        statusCode: 400,
+      });
+    });
```
Sc0ra, line 260: I think it would be better to comparer the expectedOutput with JSON.parse(result) to have a better test debug flow
Also, if it is not done in the other test cases of this file, it would be great

clementmarcilhacy: seen together, it is ok like that !

adriencaccia (approved):

adriencaccia:
@allcontributors add @clementmarcilhacy for code and docs

adriencaccia:
Fix #629

PR #633, opened 2023-04-21
Add better credentials plugin to orchestrator service

GuillaumeLagrange:
Should we also add it to the examples ? Or is it not worth it ?

image

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -853,6 +853,9 @@ importers:
       serverless-analyze-bundle-plugin:
         specifier: ^2.0.0
         version: 2.0.1(serverless@3.27.0)
+      serverless-better-credentials:
+        specifier: ^1.1.3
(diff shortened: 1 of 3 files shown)


adriencaccia (approved):
> Should we also add it to the examples ? Or is it not worth it ?

Since we do not talk of SSO in the docs, and [instead show how to create an user](https://www.swarmion.dev/docs/getting-started/get-started-on-aws/), we can leave the docs as it is.

Moreover, when we deprecate the serverless framework in favor of the CDK, this will become obsolete.

CorentinDoue:
@GuillaumeLagrange @adriencaccia FYI [better-credentials is broken for new SSO configuration](https://github.com/thomasmichaelwallace/serverless-better-credentials/issues/18) (AWS changed its config file format). 

=> New users won't be able to deploy the orchestrator stack without a manual workaround (reformat the file to legacy fomart which is still supported)

PR #648, opened 2023-05-05
fix(serverless-contracts): handle undefined fixed values in mockHandlerInput

### LINKED ISSUE #647: Omit certain fields from autogeneration in `getMockHandlerInput`
### Clear and concise description of the problem

The `getMockHandlerInput` fill up the contract's fields as intended, but some cases it would be great if we can define attributes that we do not want to fill with fake data. (The data can be in the body or in the query strings parameters...)

Why:
- For my case there is some `if` in the handler which manipulates the data according from the input. It can not 
work around every option with a different schema.

### Suggested solution

The `getMockHandlerInput` has an optional array, where can be defined which fields we do not want to populate.

For example:
```typescript
    getMockHandlerInput(myContract, {
      body:{
        name: 'John'
      },
      excludeFields:['address', 'phoneNumber']
    }
```

### Alternative

I tried the `undefined` that does not worked. Other idea was the `null`, but my case the field's value can not be `null`.

### Additional context

_No response_

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/swarmion/swarmion/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://www.swarmion.dev/docs/).
- [X] Check that there isn't already an issue that request the same feature to avoid creating a duplicate.

Sc0ra:
Fix #647

### DIFF
--- a/packages/serverless-contracts/src/contracts/apiGateway/testUtils/__tests__/mockHandlerInput.test.ts
+++ b/packages/serverless-contracts/src/contracts/apiGateway/testUtils/__tests__/mockHandlerInput.test.ts
@@ -27,8 +27,14 @@ describe('apiGateway lambda handler mock input', () => {
 
   const bodySchema = {
     type: 'object',
-    properties: { foo: { enum: ['azer', 'tyui'] }, bar: { type: 'number' } },
-    required: ['foo', 'bar'],
+    properties: {
+      foo: { enum: ['azer', 'tyui'] },
+      bar: { type: 'number' },
+      baz: { type: 'string' },
+      caboom: { type: 'array', items: [{ type: 'boolean' }] },
+    },
+    required: ['foo', 'bar', 'caboom'],
+    additionalProperties: false,
   } as const;
 
   const outputSchema = {
@@ -91,7 +97,12 @@ describe('apiGateway lambda handler mock input', () => {
 
   it('should generete event with some defined keys', () => {
     const [event] = getMockHandlerInput(httpApiContract, {
-      body: { foo: 'azer', bar: 123 },
+      body: {
+        foo: 'azer',
+        bar: 123,
+        baz: undefined,
+        caboom: [true, false, true],
+      },
       requestContext: {
         authorizer: {
           claims: {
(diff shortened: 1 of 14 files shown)


packages/serverless-contracts/src/contracts/apiGateway/testUtils/mockHandlerInput.ts
```diff
@@ -104,3 +108,14 @@ export const getMockHandlerInput = <
...
+const keepUnchangedRefsOnly = (
+  objValue: unknown,
+  srcValue: unknown,
+): unknown => {
+  if (_.isPlainObject(objValue)) {
+    return _.assignWith({}, objValue, srcValue, keepUnchangedRefsOnly);
+  }
+
+  return undefined;
+};
```
guillaumeduboc, line 121: we might want to unit test this, I'm not 100% sure what it does

packages/serverless-contracts/src/types/__tests__/constrainedJSONSchema.type.test.ts
```diff
@@ -1,8 +1,50 @@
...
+  type: 'object',
+  properties: { userId: { type: 'string' }, pageNumber: { type: 'string' } },
+  required: ['userId', 'pageNumber'],
+  additionalProperties: false,
+} as const;
+
+type SimpleSchemaCheck = A.Extends<typeof simpleSchema, ConstrainedJSONSchema>;
+
+const simpleSchemaCheck: SimpleSchemaCheck = 1;
+simpleSchemaCheck;
```
guillaumeduboc, line 21: couldn't we use a generic `Expect<T, V>` or `Assert<T>` to do our type testing instead of this ? 
you can find examples here https://dev.to/tylim88/typescript-test-your-generic-type-part-1-4jbc

packages/serverless-contracts/src/contracts/apiGateway/testUtils/__tests__/mockHandlerInput.test.ts
```diff
@@ -91,7 +97,12 @@ describe('apiGateway lambda handler mock input', () => {
 
   it('should generete event with some defined keys', () => {
     const [event] = getMockHandlerInput(httpApiContract, {
-      body: { foo: 'azer', bar: 123 },
+      body: {
+        foo: 'azer',
+        bar: 123,
+        baz: undefined,
```
guillaumeduboc, line 103: is it working if we don't specify `baz` or it should be explicitly undefined ?

packages/serverless-contracts/src/contracts/apiGateway/testUtils/mockHandlerInput.ts
Sc0ra: :magic_wand:
This is kinda tested by the unit test you commented
But yeah I could split it and do some unit tests

packages/serverless-contracts/src/contracts/apiGateway/testUtils/__tests__/mockHandlerInput.test.ts
Sc0ra: If you don't especially specify baz, a random value will be generated

guillaumeduboc (approved):
love it 🚀

adriencaccia (changes requested):
Great!
Can you add a small block in the docs to reference this change?

adriencaccia (approved):

PR #650, opened 2023-05-05
Bump nx to v16

adriencaccia:
- feat(nx-plugin): bump nx to 16
- examples: bump nx


packages/nx-plugin/package.json
```diff
@@ -73,7 +72,6 @@
     "vitest": "0.30.0"
   },
   "peerDependencies": {
-    "nx": "^15.7.0",
```
guillaumeduboc, line 76: Why do we remove the peerDep ? aren't we supposed to have nx to use the generator ?

Sc0ra (approved):

PR #651, opened 2023-05-07
feat: test nx cloud dte

Sc0ra:


.github/workflows/pr-nx-cloud.yml
```diff
@@ -0,0 +1,91 @@
...
+
+          pnpm nx run-many -t test-unit --parallel=3 &
+          pids+=($!)
+
+          # run all commands in parallel and bail if one of them fails
+          for pid in ${pids[*]}; do
+            if ! wait $pid; then
+              exit 1
+            fi
+          done
```
valentinbeggi, line 56: The simpler "stop-agents-after" and "stop-agents-on-failure" wouldn't work here ?

contracts/orchestrator-contracts/src/contracts/index.ts
```diff
@@ -1,2 +1,4 @@
+//affected
```
valentinbeggi, line 1: 🙈

adriencaccia (approved):
Great, would love to see the impact of this.
Have you got a way of getting it @Sc0ra?

Issue #657, opened 2023-05-24
Remove aws-cdk-lib as a peer dependency of serverless-helpers


RyanT5:
### Describe the bug

I was doing an investigation of our lambda bundle sizes when I discovered many of our lambdas that use parts of swarmion had aws-cdk-lib included in their deployed bundle. This adds significantly to the bundle size at 18.5mb and increases cold start times.

After reading through the swarmion repo I believe the cause to be that aws-cdk-lib is a peer dependancy in the serverless-helpers package.json. I'm not sure if there is another function in the package that requires this that I haven't seen but none of the common swarmion helper functions I've used require this so I believe it can either be fully removed or have its scope reduced to just where it is needed.

### Reproduction

- Deploy a lambda that uses only the `getEnvVariable` function from swarmion/serverless-helpers to parse a env variable in the handler
- The deployed bundle size includes 18.5mb of aws-cdk-lib
- Replace `getEnvVariable` from swarmion/serverless-helpers with a copied local function
- The deployed bundle size no longer includes 18.5mb of aws-cdk-lib

This should also work with any other serverless-helpers functions but `getEnvVariable` was tested in this case

### System Info

```shell
System:
    OS: macOS 13.3.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 68.55 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.8.0 - ~/.nvm/versions/node/v18.8.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.8.0/bin/yarn
    npm: 8.18.0 - ~/.nvm/versions/node/v18.8.0/bin/npm
  Browsers:
    Chrome: 113.0.5672.126
    Safari: 16.4
  npmPackages:
    @swarmion/serverless-contracts: 0.28.2 => 0.28.2 
    @swarmion/serverless-helpers: 0.28.2 => 0.28.2
```

### Validations

- [X] Read the [Contributing Guidelines](https://github.com/swarmion/swarmion/blob/main/CONTRIBUTING.md).
- [X] Read the [docs](https://www.swarmion.dev/docs/).
- [X] Check that there isn't [already an issue](https://github.com/swarmion/swarmion/issues) that reports the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

guillaumeduboc:
[...]

PR #660, opened 2023-05-30
fix: remove aws-cdk-lib as peer dependency

mattzcarey:
Having this as a peer dependency means that it is packaged in your lambda functions. Adds almost 20mb to your lambda package size. Not ideal :(

### DIFF
--- a/packages/serverless-helpers/package.json
+++ b/packages/serverless-helpers/package.json
@@ -63,7 +63,6 @@
     "vitest": "0.30.0"
   },
   "peerDependencies": {
-    "@serverless/typescript": ">=3",
-    "aws-cdk-lib": "^2.60.0"
+    "@serverless/typescript": ">=3"
   }
 }


guillaumeduboc:
Hey @mattzcarey,
as you can see in #657, we are not able to reproduce the issue. The package is used for importing a type so it will never appear in a bundle. Furthermore, the package can be tree-shaken and the function that imports aws-cdk-lib will be removed from your bundle

mattzcarey:
> Hey @mattzcarey, as you can see in #657, we are not able to reproduce the issue. The package is used for importing a type so it will never appear in a bundle. Furthermore, the package can be tree-shaken and the function that imports aws-cdk-lib will be removed from your bundle

Hi @guillaumeduboc. I have been doing some testing for Ryan this morning. You are very correct that this is not the issue. It still leaves us a little confused about how it solved the issue last week temporarily. 

I will close this PR.

PR #685, opened 2023-06-16
fix(orchestrator): fix dynamodb client definition

Sc0ra:


GuillaumeLagrange (approved):
lgtm to me

GuillaumeLagrange:
@Sc0ra we could open an issue on dyn db toolbox and link back to here to keep track

PR #686, opened 2023-06-16
645 rename existing serverless examples to sls

GuillaumeLagrange:
`serverless-framework-full-stack` created function names that were too long for serverless framework to handle in the package generation test (`scripts/check-example.sh`)

I went with `sls-framework`


adriencaccia:
We could even go with just `sls`.
That way we would have `sls`, `cdk` and `sst `

Sc0ra (approved):

Sc0ra (approved):

PR #794, opened 2023-11-26
feat(documentation): deploy with CDK

Sc0ra:


.github/workflows/pr.yml
```diff
@@ -51,3 +53,24 @@ jobs:
...
+      - name: Install & cache node dependencies
+        uses: ./.github/actions/install-node-deps
+      - name: Configure AWS credentials with assume role
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ secrets.AWS_STAGING_ROLE_ARN }}
+          aws-region: eu-west-1
+      - name: Deploy staging
+        # nx run-many is used here, because it will not throw an error if our target action isn't found in the project, which is exactly what we want!
+        run: pnpm nx run-many --target=deploy-staging --parallel=2
```
fargito, line 76: This is temporary I guess?

contracts/orchestrator-contracts/package.json
```diff
@@ -43,7 +43,7 @@
     "ts-node": "^10.9.1",
     "tsc-alias": "^1.8.8",
     "tsup": "^8.0.0",
-    "typescript": "^5.3.2",
+    "typescript": "5.0.4",
```
fargito, line 46: I'm not sure we should downgrade TS. Instead I think we should ditch `ts-node` which is internally used by Serverless and is no longer maintained. I'll try and push some updates to this PR

fargito (changes requested):

fargito (approved):

Issue #525, opened 2023-02-10
[RFC] Add cloud-native integration tests good practices & utils


CorentinDoue:
Hi Swarmion 🐝

To help Swarmion to be a serverless Dx heaven, I would love to add cloud-native integration tests good practices & utils in Swarmion scope.

Concretely I want to open source the code I did with @MaximeVivier to help manage the integration test environment variables. 
The interface will be something like

```ts 
// cdk stack
new TestEnvVar(this, "API_URL", { value: httpApi.url }); // => create a SSM parameter /testEnvVar/API_URL
```

```ts
// setup file of the test runner 
await syncTestEnvVars(); // get /testEnvVar/API_URL from SSM or cache and push it into process.env.API_URL
```

```ts
// integration test file

describe("health check", () => {
  it("returns 200", async () => {
    const response = await axios.get(`${TestEnvVars.API_URL}/health`);
    expect(response.status).toBe(200);
  });
});

```

The plan is:
1. Create and publish a new @swarmion/something library with the above code
2. Add an integration test example to swarmion-full-stack starter
3. add some doc to https://www.swarmion.dev/ about swarmion testing strategy

WDYT?

adriencaccia:
Hey @CorentinDoue, I really like the idea of bringing state-of-the-art Cloud Native integration tests into Swarmion 🤩 

Your plan is sound, let's start with the first step and then go on from there.
We might want to add it to the generator as well 😉

fargito:
@CorentinDoue this looks quite good, maybe `@swarmion/integration-tests` would be a good name

CorentinDoue:
Need to scope by service

type file can be committed

PR #996, opened 2024-03-06
Feat(serverless-contracts): add SQS contract

CorentinDoue:
# SQS Contract

SQS is not easy to use. Messages are serialized and batched. This contract aims to simplify the usage of SQS. It abstracts serialization and deserialization of the messages while providing type safety at runtime and in the IDE. It also abstracts technical batching process to write or read list of messages.

## Features

- `getSendMessage`: returns a function that **validates**, **serializes** and **send** a single message through SQS
- `getSendMessages`: returns a function that **validates**, **serializes**, **batches** and **send** a list of messages through SQS. Handle **retry** if the SQS is throttled because there are too many messages
- `getHandler`: returns a type safe lambda handler that **deserializes**, **validates**, **handle batched messages one by one** and **properly inform SQS of failed messages**. The callback (internal handler) receives one message and will be called in parallel for each message of the batch. If it throws, the lambda will not throw and only its message will be retuned to the SQS.
- `getLambdaHandler` (naming is bad, but is the same as AGW contract): returns a type safe lambda handler that **deserializes** and **validates** messages. The callback (internal handler) receives the whole batch of messages and is free to process them as it wants.

## TODO before merge

- [x] Publish alpha: https://www.npmjs.com/package/@swarmion/serverless-contracts/v/0.32.0-alpha.0
- [ ] e2e test the alpha on some projects
- [x] Add documentation: https://docs-preview-996--swarmion.netlify.app/docs/how-to-guides/use-serverless-contracts/sqs/

### DIFF
--- a/packages/serverless-contracts/src/contracts/SQS/utils/messagesValidation.ts
+++ b/packages/serverless-contracts/src/contracts/SQS/utils/messagesValidation.ts
@@ -1,0 +1,74 @@
+import {
+  getBodyValidator,
+  getMessageAttributesValidator,
+} from './getRecordsValidator';
+import { SQSContract } from '../sqsContract';
+import { SendMessageBuilderOptions, SqsMessage } from '../types';
+
+export const validateMessage = <Contract extends SQSContract>({
+  contract,
+  message,
+  options,
+  index,
+}: {
+  contract: Contract;
+  message: SqsMessage<Contract>;
+  options: SendMessageBuilderOptions<Contract>;
+  index?: number;
+}): void => {
+  const { body, messageAttributes } = message;
+
+  const bodyValidator = getBodyValidator<Contract>(contract, options);
+  if (bodyValidator !== undefined && !bodyValidator(body)) {
+    console.error(
+      `Error: Invalid message body ${
+        index !== undefined ? `at index ${index}` : ''
+      }`,
+      JSON.stringify(bodyValidator.errors, null, 2),
+    );
+    throw new Error('Error: Invalid message body');
+  }
+
+  if (messageAttributes === undefined) {
+    return;
+  }
+
+  const messageAttributesValidator = getMessageAttributesValidator<Contract>(
+    contract,
+    options,
+  );
+  if (
+    messageAttributesValidator !== undefined &&
+    !messageAttributesValidator(messageAttributes)
+  ) {
+    console.error(
+      `Error: Invalid message attributes ${
+        index !== undefined ? `at index ${index}` : ''
+      }`,
+      JSON.stringify(messageAttributesValidator.errors, null, 2),
+    );
+    throw new Error('Error: Invalid message attributes');
+  }
+};
+
+export const validateMessages = <Contract extends SQSContract>({
+  contract,
+  messages,
+  options,
+}: {
+  contract: Contract;
+  messages: SqsMessage<Contract>[];
+  options: SendMessageBuilderOptions<Contract>;
+}): void => {
+  const errors = [];
+  messages.forEach((message, index) => {
+    try {
+      validateMessage<Contract>({ contract, message, options, index });
+    } catch (error) {
+      errors.push(error);
+    }
+  });
+  if (errors.length > 0) {
+    throw new Error('Error: Invalid message');
+  }
+};
--- a/packages/serverless-contracts/src/contracts/SQS/utils/serializeMessageAttributes.ts
+++ b/packages/serverless-contracts/src/contracts/SQS/utils/serializeMessageAttributes.ts
@@ -1,0 +1,62 @@
+import { SendMessageRequest } from '@aws-sdk/client-sqs';
+import { MessageAttributeValue } from '@aws-sdk/client-sqs/dist-types/models/models_0';
+import { JSONSchema } from 'json-schema-to-ts';
+
+import { StringOrNumberDictionaryJSONSchema } from '../../../types/constrainedJSONSchema';
+import { SQSContract } from '../sqsContract';
+
+const serializeMessageAttribute = (
+  attributeValue: unknown,
+  attributeSchema: JSONSchema,
+): MessageAttributeValue => {
+  if (typeof attributeSchema === 'boolean') {
+    throw new Error('Invalid contract messageAttributesSchema');
+  }
+  const { type } = attributeSchema;
+
+  switch (type) {
+    case 'string':
+      return { StringValue: attributeValue as string, DataType: 'String' };
+    case 'number':
+      return {
+        StringValue: (attributeValue as number).toString(),
+        DataType: 'Number',
+      };
+    default:
+      throw new Error(
+        'Invalid messageAttributesSchema. Only string and number are currently supported',
+      );
+  }
+};
+export const serializeMessageAttributes = <
+  MessageAttributes extends StringOrNumberDictionaryJSONSchema,
+  Contract extends SQSContract,
+>(
+  messageAttributes: MessageAttributes,
+  contract: Contract,
+): SendMessageRequest['MessageAttributes'] => {
+  const { properties } = contract.messageAttributesSchema;
+
+  if (properties === undefined) {
+    throw new Error('Invalid contract messageAttributesSchema');
+  }
+
+  return Object.entries(properties).reduce<
+    SendMessageRequest['MessageAttributes']
+  >(
+    (acc, [attributeName, attributeSchema]) => {
+      if (messageAttributes[attributeName] === undefined) {
+        return acc;
+      }
+
+      return {
+        ...acc,
+        [attributeName]: serializeMessageAttribute(
+          messageAttributes[attributeName],
+          attributeSchema,
+        ),
+      };
+    },
+    {} as SendMessageRequest['MessageAttributes'],
+  );
+};
--- a/packages/serverless-contracts/vitest.config.ts
+++ b/packages/serverless-contracts/vitest.config.ts
@@ -5,6 +5,7 @@ import { configDefaults, defineConfig } from 'vitest/config';
 export default defineConfig({
   plugins: [tsconfigPaths(), codspeedPlugin()],
   test: {
+    mockReset: true,
     coverage: { reporter: ['text-summary', 'lcovonly'] },
     globals: true,
     exclude: [...configDefaults.exclude, '**/*.type.test.ts'],
--- a/user-docs/documentation/docs/how-to-guides/1-use-serverless-contracts/3-sqs.md
+++ b/user-docs/documentation/docs/how-to-guides/1-use-serverless-contracts/3-sqs.md
@@ -1,0 +1,357 @@
+---
+sidebar_position: 2
+---
+
+# Use SQS contracts
+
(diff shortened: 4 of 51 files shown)


CorentinDoue:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=swarmion&pullRequest=996) **Quality Gate failed**
> Failed conditions  [77.1% Coverage on New Code](https://sonarcloud.io/component_measures?id=swarmion&pullRequest=996&metric=new_coverage&view=list) (required ≥ 80%)
> 
> [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=swarmion&pullRequest=996)

image
image

The coverage is destroyed by the types' folder/files that contains only types. Do you want / Can you exclude them from the sonar coverage?

CorentinDoue:
@swarmion/core ⚠️ I think your dependabot is becoming crazy. I think all test (linter, types, ...) hits nx remote cache so all dependabots PR are green and it merges all of them.

For example, when I really execute the test-type command I can see that all vitest.config.ts are broken. Probably since https://github.com/swarmion/swarmion/commit/d71fd6174999ee702eb02988882488cc66396f37

image

fargito:
@CorentinDoue my first guess would be that the lockfile isn't properly taken into account by Nx to compute its cache key. I'll try and look into this

fargito:
@CorentinDoue I checked on `main` without any Nx cache and everything works as expected. I also tried to run tests on this branch with no cache and it breaks. Can you rebase your branch on main? I think it would solve the type issue

user-docs/documentation/docs/how-to-guides/1-use-serverless-contracts/3-sqs.md
```diff
@@ -0,0 +1,309 @@
...
+
+```ts
+import { getHandler } from '@swarmion/serverless-contracts';
+import { ajv } from 'libs/ajv';
+
+export const main = getHandler(mySqsContract, {
+  ajv,
+  handleBatchedRecords: false,
+})(async ({ Records }) => {
+  Records.forEach(message => {
```
fargito, line 176: Nit: the AWS API convention is a bit weird with its UpperCamelCase object keys, I would prefer to use `records`, WDYT?

user-docs/documentation/docs/how-to-guides/1-use-serverless-contracts/3-sqs.md
```diff
@@ -0,0 +1,309 @@
...
+    const { toto } = message.body; // parsed and typed with the correct keys
+
+    // write your business logic for one message
+  });
+  // Handle message failure as you want.
+  // Be aware that unhandled failure will make the whole batch of messages aivalable after the visbility timeout.
```
fargito, line 182:
```suggestion
  // Be aware that unhandled failure will make the whole batch of messages available after the visibility timeout,
```

user-docs/documentation/docs/how-to-guides/1-use-serverless-contracts/3-sqs.md
```diff
@@ -0,0 +1,309 @@
...
+  validateMessage: false,
+});
+
+await sendMyMessage({ body: { what: 'ever' } });
+```
+
+:::
+
+:::caution
+This only works with the AWS sdk v3
```
fargito, line 264:
```suggestion
This only works with the AWS SDK v3
```

swarmion.code-workspace
fargito, line 1: I think these formatting changes will cause a prettier fail, can you run `pnpm format-fix` at the repo root?

packages/serverless-contracts/vitest.config.ts
```diff
@@ -5,6 +5,7 @@ import { configDefaults, defineConfig } from 'vitest/config';
 export default defineConfig({
   plugins: [tsconfigPaths(), codspeedPlugin()],
   test: {
+    mockReset: true,
```
fargito, line 8: Curious: why is that necessary now?

packages/serverless-contracts/src/contracts/cloudFormation/__tests__/fullContractSchema.type.test.ts
```diff
@@ -2,8 +2,9 @@ import { JSONSchema } from 'json-schema-to-ts';
 
 import { FullContractSchemaType } from '../types';
 
-type Check =
-  FullContractSchemaType<'hello'> extends JSONSchema ? 'pass' : 'fail';
+type Check = FullContractSchemaType<'hello'> extends JSONSchema
+  ? 'pass'
+  : 'fail';
```
fargito, line 7: I'm a little bit confused by this change, it should disappear with a `pnpm format-fix` at the repo root

packages/serverless-contracts/src/contracts/eventBridge/__tests__/fullContractSchema.type.test.ts
```diff
@@ -9,10 +9,11 @@ const eventBridgeContract = new EventBridgeContract({
...
 
-type Check =
-  FullContractSchemaType<typeof eventBridgeContract> extends JSONSchema
-    ? 'pass'
-    : 'fail';
+type Check = FullContractSchemaType<
+  typeof eventBridgeContract
+> extends JSONSchema
+  ? 'pass'
+  : 'fail';
```
fargito, line 16: I'm a little bit confused by this change, it should disappear with a `pnpm format-fix` at the repo root

packages/serverless-contracts/src/contracts/SQS/utils/serializeMessageAttributes.ts
```diff
@@ -0,0 +1,61 @@
...
+    default:
+      throw new Error(
+        'Invalid messageAttributesSchema. Only string and number are currently supported',
+      );
+  }
+};
+export const serializeMessageAttributes = <
+  MessageAttributes extends Record<string, unknown>,
+  Contract extends SQSContract,
+>(
```
fargito, line 33: So this only supports a single depth of objects? If that is the case, maybe this should be mentioned in the docs

packages/serverless-contracts/src/contracts/SQS/utils/messagesValidation.ts
```diff
@@ -0,0 +1,74 @@
...
+export const validateMessage = <Contract extends SQSContract>({
+  contract,
+  message,
+  options,
+  index,
+}: {
+  contract: Contract;
+  message: SqsMessage<Contract>;
+  options: SendMessageBuilderOptions<Contract>;
+  index?: number;
```
fargito, line 17: Could this be required?

fargito (changes requested):
This is awesome! I have a couple questions/suggestions, but this is impressive work

user-docs/documentation/docs/how-to-guides/1-use-serverless-contracts/3-sqs.md
CorentinDoue: I agree, I changed it

CorentinDoue:
Ok, 

> @swarmion/core ⚠️ I think your dependabot is becoming crazy. I think all test (linter, types, ...) hits nx remote cache so all dependabots PR are green and it merges all of them.
> 
> For example, when I really execute the test-type command I can see that all vitest.config.ts are broken. Probably since [d71fd61](https://github.com/swarmion/swarmion/commit/d71fd6174999ee702eb02988882488cc66396f37)
> 
> image
> 
> ➡️ it breaks my PR 😢

UPDATE: It was my PR, it was started a long time ago and when I rebased it on main with the new dependencies pnpm handled pretty bad the conflicts and ended downgrading some dependencies to the version they were when I started the PR. 

➡️ I fixed this with a rebase interactive, stopped at each commit that changed the `pnpm.lock`, for each of them, I reverted the broken lock and reinstalled the dependencies I changed in my commit from the package.json

swarmion.code-workspace
CorentinDoue: I removed this commit after fixing the pnpm.lock (cf https://github.com/swarmion/swarmion/pull/996#issuecomment-2041456965)

packages/serverless-contracts/src/contracts/cloudFormation/__tests__/fullContractSchema.type.test.ts
CorentinDoue: I removed this commit after fixing the pnpm.lock (cf https://github.com/swarmion/swarmion/pull/996#issuecomment-2041456965)

packages/serverless-contracts/src/contracts/eventBridge/__tests__/fullContractSchema.type.test.ts
CorentinDoue: I removed this commit after fixing the pnpm.lock (cf https://github.com/swarmion/swarmion/pull/996#issuecomment-2041456965)

packages/serverless-contracts/vitest.config.ts
CorentinDoue: It's not necessary, but I prefer to have the mocks reset before each test. It avoids side effects of one test on another. Otherwise, lots of `beforeEach` are needed to explicitly reset the mocks when some side effects are discovered, but some can be forgotten.

It's especially useful when counting mock calls. You want all the calls of the previous test to have been forgotten

packages/serverless-contracts/src/contracts/SQS/utils/messagesValidation.ts
CorentinDoue: No, the`validateMessage` is used in both `sendMessage` and `sendMessages`. In `sendMessage` there is only one message, so there is no `index`. 

`index` is used here only to print better error messages

CorentinDoue:
@fargito I fixed the PR and acted on your comments

CorentinDoue:
[RFC] I want to e2e test the contract. Actually this could be done with integration tests. I need your advice/help on some points

1. I want to create a new CDK service `services/contracts-tests` that creates some lambdas that use each contract. Each lambda has some integration test that ensure the contract works as expected. **WDYT?**
2. I can deploy and run the test on one of my account for the development purpose but it could be cool to deploy this service on the staging env and run the test after the deployment as it was done with the integration tests of the orchestrator service. I noticed the staging deployment is broken for 5 month and the integration tests are skipped. I think I probably shut down the account where they were deployed. What do you to do with this? Do you want to remove the staging or is there a plan to get back a dedicated account and fix the CD?

fargito (approved):
Great work! Thanks a lot.
As for the integration tests:
- I agree with the new service. Maybe we can simply rename the `orchestrator` service and use its infrastructure, since it used to be the source of integration tests
- I'm not really sure we need two separate accounts for prod and staging. We could simply run the integration tests on production after merging on main

Anyway, let's continue this discussion in #525 as on a broader scale, it could also serve as a demonstration of cloud-native integration testing good practices.

fargito:
@allcontributors please add @CorentinDoue for docs

fargito:
I need to fix #1143 before releasing the beta for 0.32.0, but I'll try and do it fast

PR #1131, opened 2024-04-05
Revert "fix(serverless-contracts): update dependency tsx to v4"

guillaumeduboc:
Reverts swarmion/swarmion#1124

I believe tsx v4 is incompatible with serverless-framework
cf https://github.com/swarmion/swarmion/pull/808#issuecomment-1833392426


fargito:
Ah my bad, I was a bit too eager... We'll wait after removing the Serverless Framework then 😉

fargito (approved):

PR #1172, opened 2024-04-13
feat(serverless-contracts): control sendMessages throughput

CorentinDoue:
add throughtPut control over sendMessages

\+ 

Fix https://github.com/swarmion/swarmion/issues/1190


CorentinDoue:
Tests are broken because I added [p-queue](https://github.com/sindresorhus/p-queue) which is an ESM only package. I think it means it breaks the Swarmion contract package for all CJS users. 

What is your recommendation ? Downgrade `p-queue` to the lastest cjs version ? Or is there a trick to make it compatible ?

fargito (approved):
LGTM

PR #1890, opened 2025-04-21
Feat/add event bridge put events and improve putEvent

CorentinDoue:
## Changes:
- **EventBridgeContract - putEvent**: Allow to pass getter function for `eventBusName` instead of string
- **⚠️ EventBridgeContract - putEvent**: Throw if the event hasn't been sent instead of ignoring it. *⚠️ This is a breaking change, but previous behavior was very confusing*
- **EventBridgeContract - putEvents**: add new `putEvents` feature:
   - Automatic batch by 10 or up to the size limit
   - Throttle `PutEventCommand` API call according to quota
   - Throw if at least one event hasn't been sent


packages/serverless-contracts/src/contracts/eventBridge/features/putEvent.ts
```diff
@@ -6,19 +6,37 @@ import { PutEventBuilderArgs, PutEventSideEffect } from '../types/putEvent';
...
+      const failedEntry = Entries?.[0];
+      console.error(
+        `Failed to send this event: ${JSON.stringify(event, null, 2)}`,
+      );
+      throw new Error(
+        `Failed to send event. Error: ${failedEntry?.ErrorMessage} (${failedEntry?.ErrorCode})`,
+      );
+    }
+
+    return { failedEntryCount: FailedEntryCount, entry: Entries?.[0] };
```
fargito, line 41: This could be confusing if several entries fail. Maybe the error should at least contain all of them

packages/serverless-contracts/src/contracts/eventBridge/features/putEvents.ts
```diff
@@ -0,0 +1,137 @@
...
+import { EventBridgeContract } from '../eventBridgeContract';
+import { PutEventsBuilderArgs, PutEventsSideEffect } from '../types/putEvents';
+
+const PQueue = fixESMInteropIssue(_PQueue);
+
+const MAX_BATCH_SIZE = 256_000; // 256Kb
+const MAX_BATCH_LENGTH = 10;
+
+const defaultOptions = {
+  throughputCallsPerSecond: 400, // Minial default throughput. Some regions can have up to 10_000. See: https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-quota.html
```
fargito, line 19:
```suggestion
  throughputCallsPerSecond: 400, // Minimal default throughput. Some regions can have up to 10_000. See: https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-quota.html
```

packages/serverless-contracts/src/contracts/eventBridge/features/putEvents.ts
```diff
@@ -0,0 +1,137 @@
...
+  throwOnFailure: true,
+};
+
+const getThroughputQueueConfig = (
+  throughputCallsPerSecond: number,
+): { intervalCap: number; interval: number } => ({
+  intervalCap: throughputCallsPerSecond,
+  interval: 1000,
```
fargito, line 27: This should be defined in a separate constant, so it is clear that it is a constant, and not a magic number

packages/serverless-contracts/src/contracts/eventBridge/features/putEvents.ts
```diff
@@ -0,0 +1,137 @@
...
+});
+
+const computeEventSize = (event: PutEventsRequestEntry): number => {
```
fargito, line 30: Can you add a bit of JSDoc to explain why this is needed and the approximations you used?

packages/serverless-contracts/src/contracts/eventBridge/types/putEvent.ts
```diff
@@ -1,14 +1,19 @@
 import { EventBridgeClient } from '@aws-sdk/client-eventbridge';
+import { PutEventsResultEntry } from '@aws-sdk/client-eventbridge/dist-types/models/models_0';
```
fargito, line 2: Is this not exported from a better path, even in more recent versions?

user-docs/documentation/docs/how-to-guides/1-use-serverless-contracts/2-eventbridge.md
```diff
@@ -149,6 +150,68 @@ export const main = getHandler(
 });
 ```
 
+:::caution
+By default, the `putEvent` function will throw if it fails to send the event. You can override this behavior by passing a `throwOnFailure: false` option to the builder function. Then you have access to the processed `entry` containing the error message.
+:::
+
+:::caution
+This only works with the AWS sdk v3
```
fargito, line 158:
```suggestion
This only works with the AWS SDK v3
```

fargito (commented):
Very good job, thanks @CorentinDoue! A few minor changes, but otherwise, everything is good

packages/serverless-contracts/src/contracts/eventBridge/features/putEvent.ts
CorentinDoue: This is the put event (without "s") there is only one entry

fargito: Oh you're right, my bad

packages/serverless-contracts/src/contracts/eventBridge/types/putEvent.ts
CorentinDoue: The auto import of my IDE is fucked up. I fixed it

packages/serverless-contracts/src/contracts/eventBridge/features/putEvents.ts
CorentinDoue: I don't have lots of input, I copy-pasted this from https://github.com/fredericbarthelet/typebridge/blob/master/src/Bus.ts

fargito (approved):

PR #1920, opened 2025-05-22
Feat/multiple eventbridge handler

Yoctoboy:
# Feature

`getMultipleEventBridgeHandler` util that can receive multiple EventBridge contracts at once (example in `services/orchestrator/functions/onDeploymentRequested/handler.ts`)

## Tech

This feature was present in `v0.33.0-alpha.2` which has been used for more than a year in production on at least one Theodo project (it works great and is very useful for us)

It is directly taken from Sc0ra's branch `feat/multiple-eventbridge-handler` ([commits visible here](https://github.com/swarmion/swarmion/compare/main...feat/multiple-eventbridge-handler))

I only cherry-picked d4ab0ae9b10f28425aed52dddd1620f16fc24951 since all other commits contain version bumping changes.

## Release

Please tell me how I can help you release this properly :slightly_smiling_face: 

# Second commit

Fixes mistaken cache hits from NX which previously did not consider that `pnpm run package` could modify some outputs.

As you can see, lint/unit tests failed [here](https://cloud.nx.app/runs/9uODgoJKh9/task/orchestrator%3Atest-type) but did not fail with this change (nor did they fail locally with `--skip-nx-cache`) because cache hit/miss is computed too early. [Doc is here](https://nx.dev/reference/inputs#outputs-of-dependent-tasks)

No idea why SonarCloud fails though

### DIFF
--- a/packages/serverless-contracts/src/contracts/eventBridge/features/multipleLambdaHandler.ts
+++ b/packages/serverless-contracts/src/contracts/eventBridge/features/multipleLambdaHandler.ts
@@ -1,0 +1,70 @@
+import Ajv, { ValidateFunction } from 'ajv';
+import type { EventBridgeEvent } from 'aws-lambda';
+
+import { EventBridgeContract } from '../eventBridgeContract';
+import {
+  ContractEventBridgeEvent,
+  EventBridgeMultipleHandler,
+  SwarmionEventBridgeMultipleHandler,
+} from '../types/lambdaHandler';
+
+export type GetMultipleEventBridgeHandlerOptions =
+  | {
+      ajv: Ajv;
+      validatePayload?: boolean;
+    }
+  | {
+      ajv?: Ajv;
+      validatePayload: false;
+    };
+
+const defaultOptions = {
+  validatePayload: true,
+};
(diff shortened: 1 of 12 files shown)


packages/serverless-contracts/src/contracts/eventBridge/features/multipleLambdaHandler.ts
```diff
@@ -0,0 +1,70 @@
...
+  ) =>
+  <AdditionalArgs extends unknown[] = []>(
+    handler: SwarmionEventBridgeMultipleHandler<Events, AdditionalArgs>,
+  ): EventBridgeMultipleHandler<Events, AdditionalArgs> => {
+    const { validatePayload, ajv } = { ...defaultOptions, ...options };
+
+    let payloadValidator: ValidateFunction | undefined = undefined;
+    if (validatePayload) {
+      payloadValidator = ajv.compile({
+        anyOf: contracts.map(contract => contract.payloadSchema),
```
Yoctoboy, line 43: truc important : dans la version d'Axel, c'était `oneOf`.

